PR #1, opened 2017-09-19
Changes need to support the new interactive guides.

kinueng:
~1.  jQuery needs to be loaded before the interactive guide content.  The interactive guides use jQuery.  Without jQuery higher up in the HTML, the following error is thrown when loading the interactive guides~

~> blueprint.js:1 Uncaught ReferenceError: $ is not defined~
~>     at blueprint.js:1~

1.  Load any custom javascript and css files that are from guide repos

2.  Need to avoid overriding the `<pre>` styling used by third-party library [CodeMirror](https://github.com/codemirror/CodeMirror)  used by the interactive guides.

3.  New interactive guide layout

### DIFF
--- a/src/main/content/_includes/head.html
+++ b/src/main/content/_includes/head.html
@@ -12,6 +12,11 @@
   {% for css in page.css %}
     <link rel="stylesheet" href="{{ "/css/" | append: css | append: '.css' | relative_url }}">
   {% endfor %}
+
+  {% for css in page.guides-css %}
+  <link rel="stylesheet" href="{{ "/guides/" | append: css | append: '.css' }}">
+  {% endfor %}
+
   {% for css in layout.css %}
     <link rel="stylesheet" href="{{ "/css/" | append: css | append: '.css' | relative_url }}">
   {% endfor %}
--- a/src/main/content/_layouts/default.html
+++ b/src/main/content/_layouts/default.html
@@ -21,6 +21,11 @@
     {% for js in page.js %}
       <script src="{{ "/js/" | append: js | append: '.js' | relative_url }}"></script>
     {% endfor %}
+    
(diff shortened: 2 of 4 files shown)


gindik:
It is considered best practice to include Java script source files at the end of the body tag and not in the header tag. If the interactive guides require going against best practices, then there is something wrong with their implementation.

kinueng:
Hello @gindik 
1.  I updated the description and commit.  Instead of breaking best practice,  I added some Liquid code to load any `js` or `css` files specified by the front matter of the guides.  

2.  In `guide.css`, the `#guide_content pre` CSS selector is overriding the `.CodeMirror pre` selector.
image

Here is an snapshot of how the interactive guides are using CodeMirror to display source code.
image


3.  The interactive guide layout has been reviewed in the past and is still undergoing reviews currently.  There has not been any strong disapproval of the current interactive guide layout and there will most likely be more updates as we get closer to finalizing the interactive guides.

gindik:
1. There is no need to add duplicate code. Instead simply set the front matter values in the interactive guide to something like: "../guides/<path>"

2. Unfortunately that screen shot is out of context. The change must be reviewed in the context of the code that is being referenced but not present in the pull request. From our chat over Slack it appears the latest code of the interactive guide is here: https://github.com/steven1046/openliberty. Could you please confirm if this is the case? also, are the files within this repository in their right location? 

3. If the layout is still under review and not finalized, I would prefer to wait until it is before the change is merged into the site code that is already in production. Also, line 27 in src/main/content/_layouts/interactive-guide.html, why is the relative position CSS class used there?

kinueng:
1.  I will make those changes.  Thanks for the suggestion

2.  Yes that is the latest code for the circuit-breaker interactive guides

PR #14, opened 2017-09-29
Integrate the new microprofile circuit breaker interactive guide

kinueng:
1.  A new layout for interactive guides
2.  Update the build.sh script to clone the interactive guide common repo and the circuit breaker repo.  Add a step to move javascript files and CSS files to the _assets folder for jekyll-assets minification.
3.  Add logic to the All Guides view to look for any interactive guides.  Additionally add the bolt icon and label to distinguish that the guide is interactive vs. a static text guide.


src/main/content/guides.html
```diff
@@ -45,8 +46,12 @@ <h2 id="guide_counter_title">All Open Liberty guides ({{ guides.size}})</h2>
                         <h3 class="guide_title">{{guide.title}}</h3>
                         <p class="guide_description">{{guide.description}}</p>
                     </div>
-                    <img class="duration_clock_icon" src="/img/guide_duration_clock_icon_small.svg" alt="Duration">
+                    <img class="duration_clock_icon" src="/img/guide_duration_clock_icon_small.svg" title="Duration" alt="Duration">
```
gindik: We would need to consolidate the policy when it comes to adding tooltips via the title attribute in the site and the interactive guide. For now I would suggest not adding the title attribute on this particular file.

src/main/content/img/guide_lightningbolt.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gindik: At some point the images from the design team will be delivered optimized. In the meantime it's best to optimize them before being added to the repository. That way the repo size is not negatively and unnecessarily impacted and we don't have to add extra complexity to the build process, making it take longer than it should.

src/main/content/_assets/css/guides.css
```diff
@@ -106,10 +106,16 @@
     margin: 0;
 }
 
+.interactive_bolt_icon {
```
gindik: I would suggest using margin instead of padding.

src/main/content/guides.html
```diff
@@ -45,8 +46,12 @@ <h2 id="guide_counter_title">All Open Liberty guides ({{ guides.size}})</h2>
...
                     <span class="guide_duration">{{guide.duration}}</span>
+                    {% if guide.layout == 'interactive-guide' %}
+                    <img class="interactive_bolt_icon" src="/img/guide_lightningbolt.svg" title="Interactive" alt="Interactive">
```
gindik: Very minor change: please separate words with underscore guide_lightning_bolt.

src/main/content/_layouts/interactive-guide.html
```diff
@@ -0,0 +1,52 @@
...
+                        <div id="duration_container">
+                            <img src="/img/guide_duration_clock_icon_large.svg" alt="duration">
+                            <span id="guide_duration">{{ page.duration }}</span>
+                        </div>
+                        <p>{{ page.description }}</p>
+                    </div>
+                    <div id="blueprint_title" aria-label="">{{ page.title }}</div>
+                    {{ content }}
+                </div>
+                <div class="col-sm-12" id="page_nav_buttons" role="group">
```
gindik: While the rule in Bootstrap 3+ would fall back to the right one, it's best to always use the smallest setup: col-xs-12

kinueng:
Hi @gindik 
I have made the changes from the code review suggestions.  Please ignore the `c4cad22` commit which is from my development branch.  My development brach is not involved in this PR.

PR #21, opened 2017-10-07
add allow-uri-read attribute to asciidoc

JunyiSun:
For future guides, we can set the common-include to an url `:common-includes: https://raw.githubusercontent.com/OpenLiberty/guides-common/master` instead of using relative path`:common-includes: ../guides-common/` .

I would suggest to remove the `git clone "https://github.com/OpenLiberty/guides-common.git" src/main/content/guides/guides-common` line in the `scripts/build.sh` once the update is done for all guides.

### DIFF
--- a/src/main/content/_config.yml
+++ b/src/main/content/_config.yml
@@ -14,6 +14,7 @@ asciidoctor:
     - icons=font
     - source-highlighter=coderay
     - coderay-css=style
+    - allow-uri-read
 
 google_tag_manager: GTM-TKP3KJ7
 #google_analytics: UA-104349928-1


gindik:
Some concerns I see with replacing the relative path with absolute URLs:

 - From a performance point of view, retrieving the common files in a single clone command in the build process seems preferable. Every guide that relies on the common files would require accessing resources on the local file system (very fast and reliable). If instead remote HTTPs resource requests needs to be made for every common file in every guide, depending on the server/network speed that could potentially have an impact in the build time (and reliability).

 - If the common guide repo is renamed or moved, the impact would be minimal: just a single line update to the build.sh file. If remote URLs are used across all guides on the other hand, the impact would be greater and the update process more complex.

@NottyCode when you get a chance please let us know what are your thoughts.

JunyiSun:
@yeekangc Please share what you think. 
I do agree that this change will affect the performance of the website. Although this issue suggests that it looks like a broken link on the GitHub repo: 
https://github.com/OpenLiberty/guide-microprofile-intro/issues/9

gindik:
@JunyiSun thanks so much for the link, I'm getting a better picture now. Basically what is being described in the other issue is that when viewing the guide content not on the website, but on the GitHub, some of the links are broken because of the relative paths. If absolute URLs were used on the other hand, then they would work both on the site and on GitHub and the performance item I mentioned above would simply be a trade off. If the tests you performed confirm that with the change in place the guides look fine on both places, then I think the change is perfectly fine and the trade-off reasonable.

JunyiSun:
@gindik Thank you very much for your quick reply!

yeekangc:
Yes, there is a tradeoff here indeed.

If build performance becomes a concern, we can revisit.  Thanks!

Issue #38, opened 2017-10-16
Prevent non production environments from being crawled


gindik:
Non production environments should automatically incorporate robots.txt in order to ensure that the pages are not crawled/indexed. This is so that people using a search engine to find Open Liberty will land on the actual site and not a temporary/dev/test site.

gindik:
Note that the scope of this work should prevent any environment that is not the production one from being crawled. The issue referenced above only addresses one particular development environment. Others, like openlibertydev.mybluemix.net are still being crawled. 
With regards of the implementation, rather than adding the robot.txt file if a development variable is present, I would suggest instead adding robot.txt always unless a production variable is present. This fall-safe behavior would be more reliable, preventing future development environment instances from being crawled if they happen to be missing that variable. It would also be better aligned with the built-in production variable being used by Jekyll, which is used to control which environment instances should submit analytic information.

kevinwlau:
I agree that the robots.txt should always be added unless production variable is present. I had this change in mind but haven't gotten to it yet. Will be changing this shortly.

kinueng:
I noticed that we actually wanted `robots.txt` to be added when setting the environment to PRODUCTION.  In pull request #54, we are looking for `${ENVIRONMENT} = "DEVELOPMENT"`,  I will leave this issue opened and make a second pull request to invert the current logic for adding `robots.txt`.  For now, the end result is complete

https://openlibertydev.mybluemix.net/robots.txt
https://qa-guides.mybluemix.net/robots.txt

gindik:
@kinueng in the last comment you mentioned this issue would be kept open until the logic is updated. Looking at the code, that doesn't appear to be the case.

kinueng:
Pull request #67 is merged in `master`

PR #54, opened 2017-11-02
Add robot.txt to development environments

kinueng:
1.  `build.sh` is split into two files.  One file `build.sh` contains commands that are used to setup the Bluemix environment.  `build_jekyll.sh` contains commands specific to the `jekyll` build environment.  The motivation is to allow others to build the website using other tools besides Bluemix.

2.  For development environments, include a `robots.txt` in the website `war` so that web crawlers know to avoid indexing the development environments. 

We confirmed that the robots.txt worked using Google's tester.
https://support.google.com/webmasters/answer/6062598?hl=en&ref_topic=6061961

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -1,0 +1,3 @@
+User-agent: *
+Disallow: /
+Noindex: /
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -7,24 +7,6 @@ rvm requirements
 rvm install 2.4.1
 rvm use 2.4.1 --default
 sudo apt-get -y install nodejs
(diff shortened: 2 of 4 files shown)


gindik:
Note that the build_jekyll file in the pull request includes not only the Jekyll portion, but the Maven one as well to which Jekyll is agnostic. The file name might be a bit misleading. Perhaps a different name could be used, or the script could be split further into Jekyll and Maven sections.

kinueng:
@gindik Good point.  I'll make changes to keep things from being misleading.

kinueng:
Fixes #38

PR #67, opened 2017-11-17
Update development build process

### LINKED ISSUE #38: Prevent non production environments from being crawled

kinueng:
Fixes #38

We want to change the build logic for the steps that are done for development environments.  Instead of putting the burden on the contributor to remember to set `DEVELOPMENT` in their environment, we will assume it is a development environment.  If the build logic sees a production environment, the actions for dev only will be skipped.


kinueng:
Had to add commit 1d36fc833a6bc41fc3d3cf19f91bc6d0180d43d9 because of a problem in the string comparision logic in the bash script.

PR #72, opened 2017-11-28
Add blog about the Liberty Spring Boot Starter

tjwatson:
This can be published anytime, but it should be no later than Friday, December 1st

### DIFF
--- a/src/main/content/_posts/2017-11-27-liberty-spring-boot.adoc
+++ b/src/main/content/_posts/2017-11-27-liberty-spring-boot.adoc
@@ -9,17 +9,17 @@ author_picture: https://avatars3.githubusercontent.com/u/656475
 = Open Liberty Spring Boot Starter
 Thomas Watson <http://github.com/tjwatson>
 
-Now that the Open Liberty project is in full swing we are looking for more opportunities to enable developers to use Open Liberty in their own projects. It is very common for developers to create and deploy Spring applications on Open Liberty as a simple web application, but increasingly Spring developers are using Spring Boot as a faster, simpler way to create Spring applications. Spring Boot uses a set of maven dependencies to build the applications, these dependencies are called starters and by default Spring Boot starters use Tomcat as the embedded web container. Additional starters for Jetty and Undertow are available to override Tomcat with another container implementation. So we thought it would be interesting if you could use an Open Liberty Spring Boot starter as an alternative, just like you can already do for Jetty and Undertow.
+Now that the Open Liberty project is in full swing we are looking for more opportunities to enable developers to use Open Liberty in their own projects. It is very common for developers to create and deploy Spring applications on Open Liberty as a simple web application, but increasingly Spring developers are using Spring Boot as a faster, simpler way to create Spring applications. Spring Boot uses a set of Maven dependencies to build the applications, these dependencies are called starters and by default Spring Boot starters use Tomcat as the embedded web container. Additional starters for Jetty and Undertow are available to override Tomcat with another container implementation. So we thought it would be interesting if you could use an Open Liberty Spring Boot starter as an alternative, just like you can already do for Jetty and Undertow.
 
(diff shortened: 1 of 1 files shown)


kinueng:
Hey @tjwatson,

I changed this pull request to merge into the `development` branch.  When @lauracowen approves the blog content and merges this pull request into the `development` branch, the blog post will be published on our staging environment (https://openlibertydev.mybluemix.net/blog/).  The staging environment is where we make sure the website builds fine and the blog post is rendered as expected.

Once everything looks good, the blog post will be merged into `master` to publish on openliberty.io.

I will take care of most of the merging above.  I will reach out to you once the blog is on the staging so you can do a quick look before we push to openliberty.io.

lauracowen:
Hi, looks good in general. Some minor comments:

* 'maven' should be 'Maven'.
* 'gradle' should be 'Gradle'.
* 'pom' should be 'POM' (or `pom.xml`)
* Typo 'porjects'
* I tend to put file names in backticks so they render with grey shading: eg build.gradle as `build.gradle`, `pom.xml`
* "An exclusion is needed for each starter your Spring Boot application depends on which transitively pulls in spring-boot-starter-tomcat." is a bit awkward to read. How about: "Add an exclusion for each starter that your Spring Boot application depends on that transitively pulls in spring-boot-starter-tomcat."
* "This prevents including the spring-boot-starter-tomcat in your application." <- I know what you mean but this is also a bit awkward. Can you just say "This excludes the spring-boot-starter-tomcat from your application." or "This prevents the spring-boot-starter-tomcat being included in your application."
* "...with the following:" -> "with the following code:" or "with the following XML:"
* In fact, "For maven projects this can be done with the following:" would be better as "For Maven projects, you can do this by adding the following code:" (or XML). Ditto for the Gradle version.
* Make headings 'sentence-style'. That is change "Include Liberty Spring Boot Starter" to "Include Liberty Spring Boot starter"; "Additional Open Liberty Components" to "Additional Open Liberty components" - like in a sentence, only capitalise names and the first word. Don't capitalise every word just because it's a heading.
* "ssl" to "SSL"
* "Open liberty has a modular runtime" to "Open Liberty has a modular runtime" (capital 'L')
* "Websocket" to "WebSocket"
* "These features can be enabled by" to "You can enable these features by" (Makes it clearer who would need to do the enabling. There might be other instances where this change is also relevant - basically always write 'actively' instead of in the more 'passive' way we get taught in Science lessons at school.) :)
* "Liberty specific configuration properties" to "Liberty-specific configuration properties"
* "If not specified a temporary directory will be used." to "If not specified a temporary directory is used." - unless the present tense just doesn't make sense (it usually does work fine), use the present tense as it's clearer and easier to read. Again, there might be other instances where this change is also relevant.
* "the Liberty logs are stored under a folder" <- should that be "under a folder in" or "under the folder"?
* "Once the application stops the base directory used by liberty will be deleted." to "When the application stops, the base directory used by Liberty is deleted." - "when" rather than "once" is a bit clearer for English-as-a-second language readers; comma after "stops" makes it easier to read; capital 'L'; present tense is easier to read.

All minor things but if you could make them (and check for other instances where the same changes apply - I've not marked them all), I'm happy with this. Thanks!

tjwatson:
Thanks @lauracowen I made the requested changes in a follow on commit.  Can you merge this into development branch after a final review of the changes?

kinueng:
Hi @tjwatson,
I rendered the blog post locally with your newest commit 91632e0 and 0c4f84c.

image

kinueng:
The blog card looks like this
image

tjwatson:
@kinueng  I guess we need to update the date for the day it actually gets posted?

kinueng:
@tjwatson, I'll update the date before pushing to openliberty.io

lauracowen:
Just saw the preview - 'Starter' should be lower-case (according to the rest of the post).

Also, I forgot the blog card displays the first paragraph. Really, that card should contain the point of the post (ie that it's about setting OL as the app server in the Spring Boot starter) - currently it doesn't actually tell you what the post is about.

PR #163, opened 2018-02-23
New /guides page layout

### LINKED ISSUE #144: New all guides page design
New page design for the view that shows all the guides
reference_all guides table of contents

steven1046:
Update the all guides page to reflect the overall docs layout.
Extract the doc header css into a separate file.
Add a sticky footer to the all guides page when filtering guides so it stays at the bottom.

Fixes #144


src/main/content/_assets/css/doc-header.css
```diff
@@ -1,3 +1,159 @@
+/*******************************************************************************
+ * Copyright (c) 2017 IBM Corporation and others.
```
kinueng: We should put 2018

kinueng:
Since this work is ongoing, we will close the PR for now.  We can reuse this PR when we are ready to deliver the code.

PR #169, opened 2018-02-27
Refactor the doc header to work on all doc pages using the doc type a…

steven1046:
…nd breadcrumbs from each pages frontmatter

Removed the guide_header and made /guides and the 2 javadocs page use doc_header. 
Refactored doc_header to use the front matter from the page to show breadcrumbs and title.

- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools
- [x] https://validator.w3.org/checklink
- [x] https://validator.w3.org


src/main/content/_includes/doc_header.html
```diff
@@ -21,23 +21,31 @@
...
+                            <a href="{{ breadcrumb_page.url }}"> {{ breadcrumb_page.title }}</a>
+                        </li>
+                    {% endfor %} 
                     <li class="active"> {{ page.title }} </li>
                 </ol>
             </div>
         </div>
     </nav>
+    <div id="background_container" class="black_blue_gradient_background">
+            <div id="docs_container" class="container">
```
kinueng: Do we need `#docs_container` in the header?  If we do need `#docs_container`, I think it would be more appropriate for this to be in the _layout

kinueng: I recall we only need `#background_container` so that we get the `black_blue_gradient_background` for the background of the header.  We may even be able to rename `#background_container` to just `#header_background`.  I think "container" was once referring to an actual bootstrap container, which we are no longer using (I think?)

steven1046: #background_container is used on every other page, but we can rename it to header_background for every docs page. Check the latest commit.

src/main/content/_layouts/javadocs.html
```diff
@@ -9,6 +9,11 @@
 
     {% include doc_header.html %}
 
+    <div id="header_background" class="black_blue_gradient_background">
+        <div id="docs_container" class="container">
```
kinueng: We may be able to delete `#docs_container` and move the dark background into `doc_header.html`.  My thinking is that the dark background is now only used for the header, where as before we needed the dark background to stretch all the way down the entire page.

PR #307, opened 2018-04-27
Custom adoc links

NottyCode:
Adding support for the custom link macro for asciidoc required by feature links. The goal is that you'll be able to do feature:servlet-3.1 and it'll insert a link to the feature documentation. This is an initial implementation that works for the config doc only. Once we have decided the final url structure for these we can update so the urls can work everywhere.

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


kinueng:
In the process of using this PR for #397

PR #383, opened 2018-06-20
Add enter key support

### LINKED ISSUE #361: When searching guides, support bookmarking the search results.
#### Objective:
Support a way to update the URL with the search strings so that users can bookmark the results from their custom search.

#### Idea:
When users hit "Enter" in the search bar, update the URL with the search parameter.
Today we support the `?search=` query parameter.

#### Demo of existing behavior
image

kinueng:

PR for #361 

There is no way to bookmark the search results when filtering the list of guides.  As a minimum, we can update the URL with the `search=<text>` and `tag=key` when a user hits the enter button.

#### Preview of the new behavior using the Enter key
osearch

- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/js/guides.js
+++ b/src/main/content/_assets/js/guides.js
@@ -75,6 +75,13 @@ $(document).ready(function() {
         processSearch(input_value);        
     });
 
+    $('#guide_search_input').keypress(function(event) {
+        if(event.which == 13) { // 13 is the enter key
+            var searchInput = $('#guide_search_input').val();
+            updateSearchUrl(searchInput);
+        }
+    });
+
     var query_string = location.search;
     // Process the url parameters for searching
     if(query_string.length > 0) {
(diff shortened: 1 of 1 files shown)


steven1046:
Hi @kinueng, do you think we should update the URL as the user types? I've never pressed enter myself when filtering the guides since the guides get filtered without pressing enter.

kinueng:
@steven1046 
I did look into updating the URL as the user is actively typing in search text.  I decided I would first deliver the `Enter` key support.  Dynamically updating the URL as the user types can come in a different PR.

What I found that made it look non-trivial to do dynamically updating URL.
1. Cannot regress the browser's forward and back button behaviors
2. Need to avoid having the page refresh as you manipulate the URL

There is new HTML5 APIs that people have been using to update the URL on the fly
https://developer.mozilla.org/en-US/docs/Web/API/History_API

PR #440, opened 2018-07-11
Add 404 page

wexantis:
Created 404 page
- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


src/main/content/_assets/css/404page.css
```diff
@@ -0,0 +1,364 @@
...
+        width: 429px;
+        height: 194px;
+    }
+
+    #crashed_ship {
+        width: 429px;
+        height: 194px;
+    }
+
+    #shiplight_top, #shiplight_mid, #shiplight_bot {
```
kevinwlau: This may be personal preference, but for organizational purpose, I would use a class like `.shiplight` on these elements, and for common styling across them.

src/main/content/_assets/css/404page.css
```diff
@@ -0,0 +1,364 @@
...
+
+#shiplight_bot {
+    top: 119px;
+    left: 377px;
+    width: 13px;
+    -webkit-animation-delay: -9s;
+            animation-delay: -9s;
+}
+
+#smoke1, #smoke2, #smoke3,
```
kevinwlau: Same for the smokes, having a `smoke` class to apply common styling across them.

kinueng:
@kevinwlau and I chatted.  We have been liberal with using ID css selector because of how often we change the design of the page.  When we use ID selectors, its easier to delete HTML elements AND their styling.  If we used classes, its less obvious if other HTML elements are using the class.

PR #445, opened 2018-07-16
All Guide page redesign - categories

kinueng:

Introduce new categories layout for the All Guides page 

image

- [x] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
When you click outside of the search suggestion box, the suggestions don't disappear but they do disappear when clicking directly on the search box. Do we want to close the search suggestions when clicking elsewhere?

steven1046:
It looks good, however tabbing to the popover is not accessible and jQuery knows about this bug where the menu disappears when tabbing from the search input to the next element on the page which in this case is the popover.

PR #464, opened 2018-07-23
Draft blog post for installing features from dependencies

ericglau:
Add draft blog post for installing features from Maven dependencies.


src/main/content/_drafts/2018-07-19-installing-features-from-maven-dependencies.adoc
```diff
@@ -0,0 +1,183 @@
...
+categories: blog
+author_picture: https://avatars3.githubusercontent.com/u/29640130
+seo-title: Installing Liberty features from Maven dependencies - OpenLiberty.io
+seo-description: "Install features to the Open Liberty runtime kernel by adding dependencies in your Maven or Gradle build."
+blog_description: "Install features to the runtime kernel by adding dependencies in your Maven or Gradle build, which also provides Liberty API, SPI, and Java specification dependencies for compilation."
+---
+= Installing features from Maven dependencies in Open Liberty 18.0.0.2
+Eric Lau <https://github.com/ericglau>
+
+Open Liberty 18.0.0.2 is available in several runtime packages, including as a runtime kernel without any runtime features. With the latest releases of the Liberty https://github.com/WASdev/ci.maven[Maven] and https://github.com/WASdev/ci.gradle[Gradle] plugins, you can install runtime features simply by adding dependencies in your build. This also provides the relevant dependencies for compilation, so there is no longer any need to add Liberty API, SPI, or Java specification dependencies.
```
NottyCode: Rather than say "This also provides the relevant dependencies for compilation" I would say "The feature dependencies also provide the relevant dependencies for compilation"

src/main/content/_drafts/2018-07-19-installing-features-from-maven-dependencies.adoc
```diff
@@ -0,0 +1,183 @@
...
+
+== The runtime kernel
+
+If you're using Maven, here are the coordinates for the runtime kernel:
+[source,xml]
+----
+<assemblyArtifact>
```
NottyCode: I guess we haven't done this, but I think the kernel should be a dependency as well. Perhaps a future improvement for the plugin.

src/main/content/_drafts/2018-07-19-installing-features-from-maven-dependencies.adoc
```diff
@@ -0,0 +1,183 @@
...
+[source,groovy]
+----
+dependencies {
+   libertyRuntime 'io.openliberty:openliberty-kernel:18.0.0.2'
+}
+----
+
+== Dependencies
+
+The Open Liberty 18.0.0.2 features are available on https://search.maven.org/[The Central Repository] under the `io.openliberty.features` groupId.  When features are listed as dependencies, they will be installed by the Maven or Gradle plugin as well as provide Liberty API, SPI, and Java specification dependencies for compilation.  Features can also continue to be listed as before in the plugin configuration/attributes or in the `server.xml` file, in which case those features will also be retrieved from The Central Repository for installation.  The features that are retrieved will be stored in the local Maven or Gradle cache so that the next time you run the plugin, they won't need to be downloaded again.
```
NottyCode: Rather than say "The Central Repository" I would say "Maven Central" I think the latter is more often used.

src/main/content/_drafts/2018-07-19-installing-features-from-maven-dependencies.adoc
```diff
@@ -0,0 +1,183 @@
...
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.openliberty.features</groupId>
+            <artifactId>jaxrs-2.1</artifactId>
+            <type>esa</type>
```
NottyCode: I would expect the type to be in the bom so wouldn't need to be specified here.

src/main/content/_drafts/2018-07-19-installing-features-from-maven-dependencies.adoc
```diff
@@ -0,0 +1,183 @@
...
+            <scope>provided</scope>
```
NottyCode: I would expect this to be specified in the bom so it shouldn't be necessary here.

src/main/content/_drafts/2018-07-19-installing-features-from-maven-dependencies.adoc
```diff
@@ -0,0 +1,183 @@
...
+        </dependency>
+    </dependencies>
+----
+
+=== Gradle
+Use the `libertyFeature` dependency configuration to install features. If you are using the `java` plugin, then the `libertyFeature` configuration extends from the `java` plugin's `compileOnly` configuration to provide Liberty API, SPI, and Java specification dependencies as well.
+[source,groovy]
+----
+dependencies {
+   libertyFeature 'io.openliberty.features:jaxrs-2.1:18.0.0.2'
```
NottyCode: Is the version necessary here? I ask because the version needs to match between the kernel dependency and the jaxrs one and it means updating will require many changes in lock step. We don't need this for maven so we should be able to not need this for gradle. Perhaps we do now, but this is a possible area of improvement I think.

src/main/content/_drafts/2018-07-19-installing-features-from-maven-dependencies.adoc
ericglau: The type is in the pom but Maven still looks for a jar if it's not specified here.

src/main/content/_drafts/2018-07-19-installing-features-from-maven-dependencies.adoc
ericglau: The scope was not specified in the bom, so this is still needed.  This can be a future improvement for the bom.

src/main/content/_drafts/2018-07-19-installing-features-from-maven-dependencies.adoc
ericglau: Gradle does not natively have the ability to use boms.  It is possible with third party plugins, but that is up to the user and should be out of scope for this blog.

PR #470, opened 2018-07-24
Fix extra blue on the code_command arrows

steven1046:
Fixed the end of the code_command arrows from getting cut off.
Old:
image

New:
image

- [x] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/css/guide-multipane.css
+++ b/src/main/content/_assets/css/guide-multipane.css
@@ -185,11 +185,11 @@ header {
 }
 #guide_content .code_command > .content:before {
     top: 0px;
-    background: linear-gradient(to top right, #F1F4FE 49%, #c8d6fb calc(49% + 1px), transparent calc(49% + 2px));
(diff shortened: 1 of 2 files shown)


kevinwlau:
In short, the second gradient in the background should be the complement of the first gradient. 
First parameter should be opposite direction, 2nd parameter should complement to form 99%, theoretically leaving behind 1% "border", I believe.

kevinwlau:
Actually I think this can be accomplished using a single linear gradient.

`linear-gradient(to top right, #F1F4FE 50%, #c8d6fb 51%, transparent 0%)` for the `.content:before`

kevinwlau:
As per discussion, we noticed the percentage of the gradient came close, but didn't quite match up with the 1px border on the rest of the box. In order to fix this, we can use `calc(50% - 1px)` or similar to force a 1px width of the edge border.

steven1046:
image

Current arrow

PR #473, opened 2018-07-24
Fix scrolling up and down in the code column when there is a scroll bar

steven1046:
When there's a scrollbar in the code pane, sometimes it wouldn't let you scroll because it was using the wrong element's heights/offsets to calculate if there is a scrollbar or not. This fixes that problem.

The code column has its own scroll listener because we do not want scrolling in the code column to scroll the entire page.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/js/guide-multipane.js
+++ b/src/main/content/_assets/js/guide-multipane.js
@@ -133,15 +133,15 @@ $(document).ready(function() {
     // Handle scrolling in the code column.
     // Prevents the default scroll behavior which would scroll the whole browser.
     // The code column scrolling is independent of the guide column.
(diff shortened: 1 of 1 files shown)


kevinwlau:
More specifically, scrolling down in the code pane is now sending the event to the page sometimes.

One scenario is when the code pane has its own scrollbar and is scrolled to the bottom, further scroll will be sent to the page.

There is at least one other scenario that I'm not able to determine.

steven1046:
@kevinwlau The scenarios you described have been fixed.

PR #502, opened 2018-08-08
Intro docs design

kevinwlau:
New intro docs page

Currently with draft of the intro to microprofile


src/main/content/_assets/css/intro.css
```diff
@@ -0,0 +1,63 @@
...
+}
+
+section.banner {
+    background-color: white;
+    padding: 66px 0 73px 0;
+    border-bottom: 3px solid #d4d7e3;
+}
+
+#guide_title {
+    font-family: BunueloCleanPro-Light;
```
steven1046: I think you mean to use BunueloLight font.

src/main/content/_assets/css/intro.css
```diff
@@ -0,0 +1,63 @@
...
+    color: #1b1c34;
+    letter-spacing: 0;
+    text-align: center;
+    line-height: 35px;
+    margin-bottom: 41px;
+}
+
+#guide_meta {
+    margin: 0 auto;
+    width: calc( 95% * 0.8 );
```
steven1046: Could this be simplified to 76%?

src/main/content/_layouts/intro.html
```diff
@@ -0,0 +1,32 @@
+---
+layout: doc
+type: intro
```
steven1046: I don't see where the type attribute is used.

src/main/content/_layouts/intro.html
```diff
@@ -0,0 +1,32 @@
...
+<div id="background_container" class="gray_background">
+    <div class="centered_column">
+        <div class="row">
+            <!-- Entire guide section -->
+            <div id="guide_column" class="col-sm-12 position_relative">
+                <div id="guide_content">
+                    <!-- Guide content section -->
+                    {{ content }}
+                </div>
+                <div id="copied_to_clipboard_confirmation" style="display: none;">Copied to clipboard</div>
```
steven1046: The copy to clipboard elements aren't used in the intro guide.

src/main/content/_layouts/intro.html
```diff
@@ -0,0 +1,32 @@
+---
+layout: doc
+type: intro
+css: 
+- guide-card
```
steven1046: guide-card.css isn't used.

src/main/content/_layouts/intro.html
steven1046: What would be its future use?

kevinwlau: Checking the `page.type` for intro type pages, rather than checking `page.url` for containing `/docs/intro`

steven1046: I would remove it unless 100% certain it will be used in the future, because it could become dead code if never used or looked at again.

PR #536, opened 2018-08-17
Correct behavior for floating TOC in single column view

dmorgant:
In single column view, the sticky header and the floating TOC did not play well.  The floating TOC could be scrolled under the sticky header and was "hidden" for a bit until it hit the top of the page.  Then, the sticky header would disappear. Worked with Steven and decided to un-stick the sticky header in single column view so that it could be scrolled off the page.

The second problem was that when the floating TOC hit the top of the page and became fixed there, the document would appear to jump underneath it.  So, fixed it so that the scrolling is smoother and when the TOC becomes fixed the header for the 1st section is still visible.

- [x] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/css/guide-multipane.css
+++ b/src/main/content/_assets/css/guide-multipane.css
@@ -483,12 +483,6 @@ header {
     padding: 5px;
 }
 
-@media (max-width: 767px) {
-    header {
-        position: static; /* Make the header non sticky in mobile view */
-    }
-}
-
 @media (max-width: 1199px) {
     .guide_item {
         width: 100%;
@@ -514,6 +508,9 @@ header {
 }
     
 @media (max-width: 1170px) {
+    header {
+        position: static; /* Make the header non sticky in mobile view */
+    }
 
     #github_clone_popup_container {
         display: none;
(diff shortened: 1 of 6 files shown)


dmorgant:
What changed:

**guide-multipane.css**: Removed the media query for width 767px and placed making the sticky header static under the media query for 1170px, or single column view.   There were no other restrictions for 767px so this simplifies the implementation of our design.  In single column view the header will be able to scroll off the page as the user moves through the document.

Added a new \<div> to both **guide-multipane.html** and **iguide-multipane.html**.  This div, with class scroller_anchor, is initially created with a height of 0px and positioned at the same spot as the floating TOC, \#mobile_toc_accordion_container.   As the user scrolls down from the top of the document, and the floating TOC hits the top of the page, the floating TOC is changed to be fixed there and the rest of the page should smoothly scroll underneath it.  However, when we change the floating TOC to be position fixed, we remove it from the underlying DOM content so the "page" becomes shorter by the height of the floating TOC.  This was causing the "jump" in the guide content when the TOC hit the top of the page.  To prevent this, we can keep the content height constant by making the scroller_anchor div the same height as the floating TOC thereby replacing the floating TOC in the DOM.  When the user scrolls back to the top, we again make the scroller_anchor div a height of 0px and put the floating TOC back into the DOM. The logic for this is in handleFloatingTOCAccordion() in toc-multipane.js.

PR #562, opened 2018-08-27
Include scroll width when checking window width

dmorgant:
In widening the browser window from 2-column to 3-column I noticed that the guide content column did not change width at the appropriate time.   Also, I frequently got "stuck" with a bare column the width of the TOC, but located vertically in the middle of the page.  The cause was that the JavaScript code would check for $(window).width() being greater or less than the set twoColumnBreakpoint or threeColumnBreakpoint.  

$(window).width() does not include the width of the scrollbar present on the page.  CSS media queries react to the width of the browser window including the scrollbars. In other words: The viewport, not the window (the width of the html element).  So to have our JavaScript fire at the same time a CSS media query evaluates to true I changed all the $(window).width() to  window.innerWidth, which returns the width of the window including scrollbars.

- [x] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/js/guide-multipane.js
+++ b/src/main/content/_assets/js/guide-multipane.js
@@ -403,7 +403,7 @@ $(document).ready(function() {
     function handleSectionSnapping(event){
         // Multipane view
         if(window.innerWidth > twoColumnBreakpoint) {
-            var id = getScrolledVisibleSectionID(event);
+            var id = getScrolledVisibleSectionID();
             if (id !== null) {
                 var windowHash = window.location.hash;
                 var scrolledToHash = id === "" ? id : '#' + id;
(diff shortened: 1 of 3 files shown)


src/main/content/_assets/js/guide-multipane.js
```diff
@@ -402,8 +402,8 @@ $(document).ready(function() {
     // Slow the scrolling over section headers in the guide
     function handleSectionSnapping(event){
         // Multipane view
-        if($(window).width() > twoColumnBreakpoint) {
-            var id = getScrolledVisibleSectionID();
+        if(window.innerWidth > twoColumnBreakpoint) {
+            var id = getScrolledVisibleSectionID(event);
```
steven1046: I think you removed the event in a previous PR.

PR #582, opened 2018-09-04
Fix header id mistmatch

steven1046:
When storing the id's of each code section on the right, the way we looked for the id of the section before did not match the way we checked what section a hotspot belonged to. This will create a common function that they use so they won't be mismatched again. 
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
@andrewdes The fallback guide should be fixed now.

PR #637, opened 2018-09-14
Do not open until October! Re-introduce "refactoring one" at EE 7 level

barecode:
This re-introduces the 'refactoring one' Java changes for the back-end builds API, but importantly does so at Java EE 7 API levels. The original code change used a function `Json.createArrayBuilder(collection)` which was introduced in jsonp-1.1. With a 4-line write, this now uses functions available in jsonp-1.0 which does not require any changes to the server's feature set.

Let's target merging this the first week of October.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,12 +98,7 @@
 			<plugin>
 			    <groupId>io.openliberty.boost</groupId>
 			    <artifactId>boost-maven-plugin</artifactId>
-			    <version>0.1</version>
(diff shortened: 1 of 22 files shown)


barecode:
@kinueng , @steven1046 - have a look at this when you can. The code changes are big but unlike the previous refactoring it does not introduce any new features. I've tested this on an EE 7 / MicroProfile 1.0 implementation and its good and happy.

PR #742, opened 2018-10-22
Update minification

### LINKED ISSUE #73: Use the latest jekyll-assets version
In issue #70, I had to downgrade the version of `jekyll-assets` to fix a hard build break.  We need to figure out whether this is an issue we need to raise in `jekyll-assets` or whether this is our build environment.

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


kinueng:
We will need to revisit this pull request.  We are encountering two build problems.

1.  
```
Dependency Error: Yikes! It looks like you don't have octopress-minify-html or one of its dependencies installed. In order to use Jekyll as currently configured, you'll need to install this gem. The full error message from Ruby is: 'Unable to activate multi_js-0.1.0, because uglifier-4.1.19 conflicts with uglifier (~> 2)' If you run into trouble, you can find helpful resources at https://jekyllrb.com/help/!
```

After deleting uglifier v4...a second problem appears
```
       Jekyll Feed: Generating feed for posts
  Liquid Exception: Invalid option: harmony in /_layouts/default.html
jekyll 3.8.4 | Error:  Invalid option: harmony
```

kinueng (dismissed):
Still some build breaks needing to be fixed for this PR

kinueng:
A few things:
1.  `jekyll-minifier` has a GPL v3 license that requires us to get legal approval first before we can use `jekyll-minifier`.
2.  I am having trouble disabling `jekyll-minifier` in our development environment.  I have added the snippet below in https://github.com/OpenLiberty/openliberty.io/blob/master/src/main/content/_dev_config.yml.  When added, the /docs/ has missing CSS styling...

```
jekyll-minifier:
  exclude: '**.*'
```

steven1046:
In this pr, we will also want to remove the hardcoded `jekyll` gem version if all of the latest versions can work together.

steven1046:
Some notes:
Jekyll 4.0 doesn't work with Jekyll-assets 3.0.12 so we had to keep using jekyll 3.8.6.
I couldn't get all of the ruby gem versions to build together (even after compatibility). To get past a few errors, I had to change sprockets to version 3.7 and uglifier to 4.1 but jekyll serve still fails with no helpful error with with --trace and --verbose mode on.

kinueng:
Closing this pull request and can revisit this when moving Jekyll up to a higher version.

PR #750, opened 2018-10-24
Add check to ensure that prevSection is not undefined

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/js/common-multipane.js
+++ b/src/main/content/_assets/js/common-multipane.js
@@ -164,13 +164,15 @@ function checkForInertialScrolling (event){
             // Scrolling up
             // Check to see that the current section's top is in viewport and at least 200 pixels from the top of the screen but not more than 400.
             // Scroll up by a full page's height so that the previous section ends at the bottom of the viewport for optimal reading.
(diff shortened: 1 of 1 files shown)


src/main/content/_assets/js/common-multipane.js
```diff
@@ -164,7 +164,7 @@ function checkForInertialScrolling (event){
             // Scrolling up
             // Check to see that the current section's top is in viewport and at least 200 pixels from the top of the screen but not more than 400.
             // Scroll up by a full page's height so that the previous section ends at the bottom of the viewport for optimal reading.
-            if(top > 200 && top < 400){
+            if(top > 200 && top < 400 && prevSection !== undefined){
```
steven1046: I don't think it fixes it because prevSection wasn't defined before this check and it will never go into this if statement

steven1046: I believe the error is caused by getting the previous section of the first section, which doesn't have a previous section

PR #759, opened 2018-10-31
SEO description fallback to page description

steven1046:
Add a default description for meta tags when no seo-description is provided and the page has a description.

Title will be seo title, page title appended by site title, or fallback to the site title only in that order.
Description will be seo description, page description, or fallback to the site description only in that order.

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_includes/head.html
+++ b/src/main/content/_includes/head.html
@@ -1,11 +1,19 @@
-<head>
+{% assign description = '' %}
+  {% if page.seo-description %}    
+    {% assign description = page.seo-description | escape %}
+  {% elsif page.description %}
+    {% assign description = page.description | escape %}
+  {% else %}
+      {% assign description = site.description | strip_html | normalize_whitespace | truncate: 160 | escape %}
+{% endif %}
 
+<head> 
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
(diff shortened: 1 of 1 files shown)


steven1046:
@kinueng the requested changes have been made.

We did not update the Twitter title to use the same title because we don't want to append the site title to the page title as it may be too long.

src/main/content/_includes/head.html
```diff
@@ -1,11 +1,19 @@
-<head>
+{% assign description = '' %}
```
kinueng: Please format the tabs in this block of new code.  The alignment of the code is not lined up and can be confusing what line is nested in what code block.

src/main/content/_includes/head.html
```diff
@@ -1,11 +1,19 @@
...
+  {% if page.seo-description %}    
+    {% assign description = page.seo-description | escape %}
```
kinueng: I'm going a bit back and forth with myself whether we need to `escape` our descriptions.  On one hand, `escape` will make our descriptions less human readable.  On the other hand, `escape` is good security practice for special characters.  @steven1046 your thoughts?

I also, oddly, could not actually get `escape` to work.  When I inspect the description, the special characters are not "encoded".  I used my local test environment and added characters `??` to `:page-seo-description:` to one of the guides.  I inspected the HTML and still see the non-escaped `??` in the `description` 

image

steven1046:
@kinueng is there anything else to change for this pull request since we want to keep the liquid escape?

PR #769, opened 2018-11-12
Change height to ensure all classes are visible

ellenwyllie:
- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
Tested locally and it works! I think we should also update the height used in:
function setPackageContainerHeight() {
    var packageContainer = $('#javadoc_container').contents().find(".leftBottom");
    if (packageContainer.css("height") !== "70%") {
        // restore the height in case it is collapsed
        packageContainer.css("height", "70%");
    }
}
to be 55% as well since i think it is used for the same calculation when using the back/forward arrow in the browser.

Issue #776, opened 2018-11-12
Table of contents highlight is not centered - Medium priority


justineechen:
<img width="277" alt="screen shot 2018-11-12 at 1 18 26 pm" src="https://user-images.githubusercontent.com/18709242/48367044-8c109e80-e67d-11e8-8766-26a3aaf24188.png">

PR #778, opened 2018-11-12
Fix symmetry in the TOC/Guide columns

steven1046:
Fix symmetry of left and right padding/margin in the toc and guide columns.

Made the left/right padding/margin/border equal in the multipane columns.
As you can see, the right padding was overflowing into the code column before. With the new changes the content will be within the column's boundaries.
Guide column before:
image
After:
image

The TOC also had some mismatched padding/margins which were changed to be equal on both sides as well as adding a transparent border-right equal to the one on the left that becomes orange so the spacing is equal.

- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/css/guide-multipane.css
+++ b/src/main/content/_assets/css/guide-multipane.css
@@ -236,10 +236,8 @@ header {
 
 #guide_content div.sect1 {
     padding-bottom: 50px;
-    padding-left: 60px;
-    padding-right: 60px;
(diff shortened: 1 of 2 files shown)


steven1046:
Fixes #776

PR #779, opened 2018-11-13
Allow redirects to 3rd party blogs

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


kinueng:
We still need to implement a way to redirect the hidden page created by the `adoc` file that contains the redirect URL information.  Jekyll creates a openliberty.io page for every `adoc` file, even if we choose to only use the `adoc` file to symbolize an external URL.

PR #787, opened 2018-11-14
Config overview fixes

NottyCode:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/config/serverConfiguration.adoc
+++ b/src/main/content/config/serverConfiguration.adoc
@@ -25,7 +25,7 @@ This example `server.xml` configures the server to support Java Server Pages 2.3
 to listen to incoming traffic to localhost on port 9080, and to automatically
 expand WAR files when they are deployed.
 
-The `server.xml` is described in more detail [below](#server.xml). When
+The `server.xml` is described in more detail <<server-xml,below>>. When
 discussing config for {projectName}, the term server config maybe used to refer
 to all of the files that make up the server config, or specifically to the
 config held in XML files; this is usually clear in context and, if not, the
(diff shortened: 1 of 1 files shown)


src/main/content/config/serverConfiguration.adoc
```diff
@@ -25,7 +25,7 @@ This example `server.xml` configures the server to support Java Server Pages 2.3
 to listen to incoming traffic to localhost on port 9080, and to automatically
 expand WAR files when they are deployed.
 
-The `server.xml` is described in more detail [below](#server.xml). When
+The `server.xml` is described in more detail <<server-xml,below>>. When
```
kinueng: Unfortunately, the way the server config documentation page is coded, this Asciidoc `Internal cross references` [1] does not work.  I tried to triage the malfunction and found that it is a non-trivial change to support the cross reference to `server.xml`.  I'll have to open an issue to look further.  For now, we will just have to change `<<server-xml,below>>` to `server-xml`.

[1] https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#links

NottyCode: Do you mean we can’t cross link, or we can’t cross link with a the link text? It doesn’t make sense to end with server.xml if it doesn’t link to anything. We could try an external format link, is that more likely to work?

kinueng: I'm currently adding fixes to suppose the `<<server-xml,below>>`.  The problem is that the browser scroll too far because the top navigation bar height is somehow being ignored.

Current behavior after clicking `server-xml`
image

After deleting the navigation bar...you can see where the `#server-xml` anchor is positioned too far at the top.
image

PR #832, opened 2018-12-12
Support multiple files from each subsection

steven1046:
Enable the multipane guide to support having more than one file in a subsection. It will load all of the files into tabs and the hotspots will switch to the correct file when hovering. Clicking on tabs will still work of course. When scrolling, there isn't a good way to detect what file to switch to when scrolling into a subsection with multiple files so it will show the first one and they can click on another tab or a hotspot will switch to the other file.

If more than one file is in a subsection in the guide's .adoc file, all of the hotspots above where the file is placed in the subsection will map to that file. Guide authors can override this behavior by putting [hotspot=14 file=1] for inline hotspots to specify that the hotspot should belong to the second file in that section.
- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/js/guide-multipane.js
+++ b/src/main/content/_assets/js/guide-multipane.js
@@ -30,7 +30,8 @@ $(document).ready(function() {
             var header = get_header_from_element(code_block);            
             header.setAttribute('data-has-code', 'true');
             var code_section = {};
-            code_section.code = duplicate_code_block;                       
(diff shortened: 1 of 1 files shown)


steven1046:
@lavenal I used the cdi intro guide to test multiple files.

PR #835, opened 2018-12-12
Prevent any no_copy blocks from becoming green

steven1046:
Sometimes including a file and adding role='command' to it will influence the other blocks around it. This is to add code to prevent no_copy blocks from turning green like commands. I confirmed with Justine that no_copy blocks will always be gray.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/css/guide-multipane.scss
+++ b/src/main/content/_assets/css/guide-multipane.scss
@@ -158,7 +158,7 @@ header {
     padding: 30px;
 }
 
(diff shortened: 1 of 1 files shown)


steven1046:
E.g.
image
The mvn commands using the role='command' affected the test block below it.

PR #903, opened 2019-01-21
Support [hotspot] to switch to a file but not highlight it

steven1046:
This is to support [hotspot] for hotspots to switch to a file but not highlight it.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/js/guide-multipane.js
+++ b/src/main/content/_assets/js/guide-multipane.js
@@ -239,10 +239,20 @@ $(document).ready(function() {
     // Parse the hotspot lines to highlight and store them as a data attribute.
     $('code[class*=hotspot], span[class*=hotspot], div[class*=hotspot]').each(function(){
         var snippet = $(this);
-        var code_block = get_code_block_from_hotspot(snippet);
         var classList = this.classList;
         var line_nums, ranges;
+
+        // Find if the hotspot has a file index set to override the default behavior.
         for(var i = 0; i < classList.length; i++){
+            if(classList[i].indexOf('file=') === 0){
+                var fileIndex = classList[i].substring(5);
+                snippet.data('file-index', parseInt(fileIndex));
+            }
+        }
+
+        var code_block = get_code_block_from_hotspot(snippet);
+
+        for(i = 0; i < classList.length; i++){
             var className = classList[i];
             if(className.indexOf('hotspot') === 0){
                 var fromLine, toLine;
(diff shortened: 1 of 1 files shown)


steven1046:
Need to talk to YK's team. To switch to a tab currently the guides specify 1-max line and it's fragile since if the files change then the hotspots change. The proposal is to switch to just `hotspot` instead of `hotspot=1-max lines` to switch to a file and not highlight the code.

src/main/content/_assets/js/guide-multipane.js
```diff
@@ -334,29 +345,29 @@ $(document).ready(function() {
...
                     var lines = ranges[i].split('-');
                     if(lines.length === 2){
                         var fromLine = parseInt(lines[0]);
                         var toLine = parseInt(lines[1]);
                         var num_Lines = parseInt(code_block.find('.line-numbers').last().text());
-                        if(fromLine && toLine){              
+                        if(fromLine && toLine){
                             // If a hotspot refers to a whole file, do not highlight it.    
-                            if(!(fromLine === 1 && toLine === num_Lines)){               
+                            if(!(fromLine === 1 && toLine >= num_Lines)){
```
kinueng: We discussed offline about `toLine >= numLines`.

I found it odd that the guide author would annotate their `adoc` file with a line range greater than the file's max number of lines.  It was discovered that guide authors used a line range of `1 - max line` as a way make the view focus on the code file, if the code file is in a background tab.  We also avoided highlighting the whole file even though the hotspot range is 1 through the end of the file.  Hotspot with range should mean highlight lines, even if they put a range of 1 through file's max lines.

Given that we are introducing `[hotspot]` with no range to indicate to bring the whole file into view if it is in a background tab AND not highlight any lines, this `if(!(fromLine === 1 && toLine >= num_Lines)){ ` check will no longer be needed.

We also now have a new `if(ranges)` guard above, so the only time we use this code is when the author wants to apply highlighting to the code lines via a hotspot with line range (ie `hotspot=3-60`).

steven1046:
Currently we specify the full range of code in a hotspot or hotspot=* to switch to a file but not highlight it. We should start using just hotspot to switch to a file but not highlight it, but it would require updates to the asciidocs. Visually there would be no change but it would make more sense from an author's point of view.

PR #932, opened 2019-01-28
Improve guide breakpoints

steven1046:
One of our checks for the browser width wasn't working on Safari only (it said it was in 3 columns when it clearly was only in 2 column view via css), and I found a better way to check the media query in javascript. I also updated our breakpoint checks to not be hardcoded everywhere to use common functions as well as change the max-width media queries because you can't have max-width: 1170px and min-width: 1170px or they'll both take effect in the same browser width. See the motivation for .98px here: https://stackoverflow.com/a/51567792/5352346.

This pull request depends on https://github.com/OpenLiberty/openliberty.io/pull/925 to be delivered.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
This pull request is a little messed up since it was built on the refactor of the single pane guide layout and that one was modified heavily after building this pull request. I will close this until it is ready again.

PR #964, opened 2019-02-06
Fix code block padding in blogs

kinueng:

When the blogs were upgraded to bootstrap4, the code blocks post their padding.  The padding came from the bootstrap3 source code.  Add the padding back to the code blocks in the blog posts.

I also took the time to convert two CSS to Sass

#### Before
image

#### After
image

- [x] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/css/blog.scss
+++ b/src/main/content/_assets/css/blog.scss
@@ -18,15 +18,28 @@
 
 #blog_container {
     padding-top: 190px;
+
+    @media (max-width: 767px) {
+        padding-top: 60px;
+    }
 }
 
 #blog_row_1 {
     margin-bottom: 117px;
+
+    @media (max-width: 767px) {
+        margin-bottom: 40px;
+    }
 }
 
 #page_intro_paragraph {
     font-size: 20px;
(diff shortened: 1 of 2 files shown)


src/main/content/_assets/css/blog.scss
```diff
@@ -18,15 +18,28 @@
 
 #blog_container {
     padding-top: 190px;
+
+    @media (max-width: 767px) {
```
steven1046: We should be using max-wdith 767.98px from now on. See https://getbootstrap.com/docs/4.0/layout/overview/.

src/main/content/_assets/css/blog.scss
```diff
@@ -51,6 +68,20 @@
     padding-left: 72px;
     padding-right: 70px;
     padding-bottom: 64px;
+
+    @media (max-width: 991px) {
```
steven1046: Same as above, we should use 991.98px.

PR #971, opened 2019-02-07
Transparent nav

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


ellenwyllie:
There were a few small tweaks to the hamburger menu design that can be found here: https://ibm.ent.box.com/folder/66336622995

steven1046:
The logo and hamburger look really close to the sides of the screen in tablet view because of the following css,
@media (max-width: 991.98px)
.navbar-expand-lg>.container, .navbar-expand-lg>.container-fluid {
    padding-right: 0;
    padding-left: 0;
}
In my opinion it looks a lot better if we don't remove the left/right padding, @cajaygle here is before and after:
image
image

At least for the right side, because the hamburger is overlapping the scrollbar.

src/main/content/_includes/head.html
```diff
@@ -26,11 +26,11 @@
...
   {% css coderay-custom %}
 
-  {% for css_name in page.css %}
+  {% for css_name in layout.css %}
     {% css {{css_name}} %}
   {% endfor %}
-  
-  {% for css_name in layout.css %}
+
+  {% for css_name in page.css %}
```
steven1046: Were these flipped so that one of the css files could override another?

src/main/content/_assets/css/header.css
```diff
@@ -1,13 +1,27 @@
...
+
 .navbar {
     padding: 10px 0px 10px 0px;
     width: 80%;
     margin: 0 auto;
+    z-index: 9999;
+}
+
+.bg-custom {
+    background-color: transparent !important;
```
steven1046: Is the bg-custom with !important needed with the negative margin on the <main> to make the navbar transparent?

steven1046:
Those are the only 2 comments I had, otherwise it looks good! The rest of the org can comment on the rest of the design in Slack.

src/main/content/_includes/head.html
ellenwyllie: I flipped these so that the header.css would load before the css for the individual page. That way the specific page's css would override the general css from header.css. This was so that I could customize the nav link colors on the support and downloads pages.

src/main/content/_assets/css/header.css
ellenwyllie: Hmm I don't think that's needed so I can remove that.

kinueng:
uh oh, I caused the conflict in `blog.css` because I converted it to Sass.

kinueng:
I fixed the `Transparent_Nav` branch and verified the change to `blog.scss` is still there `margin-top: 46px;`

PR #1180, opened 2019-05-07
Remove extra scroll bar, make toc fixed position

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
image
I don't think #toc_column needs overflow: scroll as it adds a white bar on the right of the toc, and #toc_inner could use overflow-auto so it only shows the scrollbar when the browser is too short to fit all of the toc in it.

PR #1190, opened 2019-05-10
Fix check for undefined

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


kinueng:
I have been trying to learn, on the fly, how all our hotspot code works.  There are three guides that have broken right code columns due to a bug in our hotspot handling code.  

The variable `code_sections`[1]  is suppose to store all the programming files needed by the guide.  We use `code_sections` to highlight lines in certain files (files are referenced by a index number in `code_sections`).  Somehow the `code_sections` variable is not completely loaded with all the files referenced by the guide and we get a console error [2]

[1] https://github.com/OpenLiberty/openliberty.io/blob/c36831e21b0064fa49ccac07d9588b482c58c75b/src/main/content/_assets/js/guide-multipane.js#L12

[2]
image

Issue #2, opened 2017-09-21
Improve site performance (consider minify resources)


kinueng:
Website performance can be improved by minifying CSS, HTML, and JavaScript files.  

[Google PageSpeed ](https://developers.google.com/speed/pagespeed/ ) results showing how much improvements can be gained if files are minified.

Results:
https://developers.google.com/speed/pagespeed/insights/?url=openliberty.io
Current values at this time: mobile 59, desktop 76

Jekyll plugin that may aid in compressing files:
http://jekyll.github.io/jekyll-assets/

gindik:
@kinueng thanks for the links / suggestions. We actually do have some of the performance items lined up for the implementation, I'll incorporate the ones above.

gindik:
Adding a second performance benchmark in order to better measure impact of changes before/after implementation. https://tools.pingdom.com/#!/cyj8qP/https://openliberty.io/

gindik:
<img width="990" alt="screen shot 2017-09-22 at 1 07 43 pm" src="https://user-images.githubusercontent.com/17207930/30756173-8aff23f0-9f97-11e7-84a5-6038807cbd38.png">

gindik:
I looked into Jekyll assets. It's definitely something we want to incorporate into the site. However, it does not appear to handle HTML minification. For that we'll have to resource to a different tool such as https://github.com/octopress/minify-html.

gindik:
I tested octopress html minify and it works fine in the development environment. However, when pushing it to production it's not working right. After trying several times I found that other people also reported the same issues: https://github.com/octopress/minify-html/issues/13. For now I'll remove the dependency and look for other ways to achieve HTML minification.

gindik:
First optimization steps: minify CSS, JS and compressed largest images (that do not require transparent background) to JPEG. Results:
 - Page size went from 583.2kB  to  499.2kB
 - Load time went from 1.34s to 1.25s
 - Site went from 83% to 85% faster than tested sites

gindik:
Further optimization:
 - Optimized SVG images (combined with the previous compression optimization, images in the site went from 890KB to 582KB)
 - Replaced centralized image cache (preloading to avoid flickering) with a distributed one.
 - Added HTML minification

Impact so far:
 - Page size went from 499.2KB to 459.1KB (originally was 583.2KB)
 - Load time went from 1.25s to 1.23s (originally was 1.34s)

gindik:
Caching is in place (for images)

[...]

PR #1192, opened 2019-05-17
Config scrolling

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
@lavenal could you please review this pr since it touches the breadcrumb code?

steven1046:
@ellenwyllie could you please look into why the breadcrumb is no longer showing up (on a long config page like application)?

steven1046:
Hi @ellenwyllie, the config page looks at lot better now but when you click on a sub-toc it scrolls to the new anchor and changes the breadcrumb while scrolling then it vanishes. The current implementation on openliberty.io scrolls to the new anchor but blocks the update of the breadcrumb during scrolling. Are we able to fix that in the new implementation?

steven1046:
2 issues i found which already exist and we may want to open issues to address them (i think we should just to talk about them):
1. clicking on a 2nd level toc item will scroll to the right spot on the page but not show up in the breadcrumb.
2. if you scroll down a config like application and click on the first breadcrumb to take you to the top of the page and click the browser refresh, there's a blank div that appears which is where the breadcrumb goes but no content is in it.

steven1046:
I think we'll need to fix #2 from https://github.com/OpenLiberty/openliberty.io/pull/1192#issuecomment-516137046 since it works on the site before the changes.

PR #1218, opened 2019-06-11
Add prereqs

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
The assumption is that the prereqs file in guides-common will have the version required in the name of the prerequisite.

steven1046:
Before delivering we should ensure that the prereq file is delivered to guides-common. (Ideally in master too since the ui code will make its way into master and needs that file in master).

PR #1341, opened 2019-08-27
Allow guide toc to be scrolled

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/css/toc.scss
+++ b/src/main/content/_assets/css/toc.scss
@@ -120,6 +120,9 @@
     }
 }
 
+#toc_inner {	
+    padding-top: 20px;	
+}
 
 #toc_indicator {
     z-index: 5;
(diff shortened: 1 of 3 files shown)


src/main/content/_assets/css/toc.scss
```diff
@@ -487,7 +488,8 @@
     }
 
     #toc_inner {
-        width: inherit;
+        height: calc(100vh - 101px);
+        overflow: auto;
```
steven1046: We may want to change it to overflow-y: auto only since on Safari it has a horizontal scrollbar at full resolution:
image

steven1046:
The horizontal scrollbar still appears even though the user can see the TOC width wise. I think we should set overflow-x: hidden so it doesn't appear and if they shrink the browser width wise it will switch into the tablet/mobile views so they'll always have enough space to see the full width.

PR #1490, opened 2020-01-09
DAP accessibility updates for the Home page

dmorgant:
What was fixed?  DAP accessibility issues on Site Home page.
Were the changes tested on
- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/css/home.scss
+++ b/src/main/content/_assets/css/home.scss
@@ -370,6 +370,7 @@ body {
         display: block;
         padding: 18px 0px;
         color: #3F4664;
+        width: 100%;
     }
 
     & .long_title {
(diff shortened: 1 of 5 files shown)


ellenwyllie:
I think it would be nice if we could make the links take up the full size of the boxes. Some of the ones with shorter text, like CDI, have a lot of space in the box that isn't clickable.

ellenwyllie:
Also, if I’m being super picky, I think the boxes could be slightly wider at certain screen sizes. Increasing the width from 68px and 70px and 85px to 87px in the media queries makes it match up better with how it currently looks on the site.

PR #1502, opened 2020-01-20
DAP accessibility updates for the Docs page

dmorgant:
- [] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/docs.html
+++ b/src/main/content/docs.html
@@ -48,23 +48,23 @@ permalink: /docs/
 
                 <div class="row">
                     <div class="col-sm-6">
-                        <a href="/docs/ref/config/" class="reference_item" aria-label="Open Liberty configuration documentation">
-                            <h3 class="reference_title">Server config</h3>
+                        <a href="/docs/ref/config/" class="reference_item" aria-label="Open Liberty server configuration documentation">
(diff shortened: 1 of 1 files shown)


src/main/content/docs.html
```diff
@@ -48,23 +48,23 @@
 
                 <div class="row">
                     <div class="col-sm-6">
-                        <a href="/docs/ref/config/" class="reference_item" aria-label="Open Liberty configuration documentation">
-                            <h3 class="reference_title">Server config</h3>
+                        <a href="/docs/ref/config/" class="reference_item" aria-label="Open Liberty server configuration documentation">
+                            <h3 class="reference_title">Server configuration</h3>
```
ellenwyllie: We probably want to keep the h3 text as is and just change the aria-label to match.

dmorgant: The problem is that the page we go to when selected is titled 'Server configuration', not 'Server config'.   They should be the same for both accessibility and NLS reasons.   Other languages may not equate 'Config' and 'Configuration'.

Issue #1242, opened 2019-06-27
Site Accessibility


mbroz2:
Tackle in this order

- [x] Home page
 All non guides/docs pages - (downloads, blogs, community, support, etc)
- [x] Downloads 
- [x] Blogs
- [x] Community
- [x] Support
- [x] Contribute
- [x] Guides
   - [x] Guide - non interactive
   - [x] Guide - interactive
- [x] Docs (keep in mind we are moving to antora)
- [x] Reference - subpages
   - [x] Java EE API - No DAP Issues
   - [x] MicroProfile API - No DAP issues
   - [x] Server Config
   - [x] Server Features
   - [x] Server Commands
   - [x] General

Also, tackle DAP, tabbing, then Jaws
- [x] DAP
- [x] JAWs
  
For any text/image that is part of logo name that has DAP violation please see https://github.com/openliberty/openliberty.io/issues/1242#issuecomment-576752714

dmorgant:
Pages without any updates:

- Community page had no DAP updates

aknguyen7:
- Contribute page has 1 DAP violation on Open Liberty text but it does not need to meet this requirement since it's a text that is part of logo

![image](https://user-images.githubusercontent.com/30509453/72820982-76114d00-3c35-11ea-84ea-e789a8e40cc9.png)


![image](https://user-images.githubusercontent.com/30509453/72821052-904b2b00-3c35-11ea-9aef-641be4daf49e.png)

mbroz2:
Can we add a comment into the html that summarizes the above comment so that anyone running DAP in the future that gets that error immediately sees that it's not an issue and doesn't have to spend any time investigating further?

dmorgant:
Accessibility and the NavBar in the header -  getting DAP color contrast issues when the NavBar elements in the header appear over one of the dark blue-ish backgrounds.

The bootstrap \<nav> element is defined with a transparent background.  The NavBar menu links are defined as children of the bootstrap \<nav> element.  Below the \<nav> element in the html is another \<div> element that the \<nav> element is sitting on top of.  This \<div> element is given a background-color or background-image that we see (dark blues).  When DAP processes, it reads the text color and the background color of the \<nav> element for the navigation links and is falsely reporting a color contrast issue because it is not comparing the color of the text with the actual color of the background the links are sitting on top of. 

FYI...setting the text color of the navigation links to black will cause these contrast issues to go away, but visibly the text is near impossible to read.

[...]

PR #1523, opened 2020-01-23
fix DAP errors for individual blog page that is part of the framework

aknguyen7:
- [x] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


aknguyen7:
1. work with Melissa Denby designer to ok the color text change to meet the contrast ratio requirement for the post_date text color.
2. fix the duplicate id for the arrow img, DAP doesn't like this
#1242

ellenwyllie:
Screen Shot 2020-01-23 at 2 25 03 PM

The second arrow is no longer the correct size because the id was changed. You could change both of the arrow ids to be classes since there are 2 on the page. That way the css will still apply for both arrows.

aknguyen7:
#1242

Issue #1018, opened 2019-02-22
Pre-populate download page cache in builds


kinueng:
The downloads page has a caching mechanism on the backend.  The cache is populated on the first load of the downloads page.  We should do this in the build rather than a human visiting the downloads page.

We should be visiting the downloads page for EVERY `cf` application.  Each geo region is backed by two `cf` apps behind one domain.  We need to somehow visit the downloads page for individual applications during the build to populate the cache in each application.

mbroz2:
Are we unable to just make a server side call to DHE during server/app startup to populate the cache?

steven1046:
@kinueng @mbroz2 What is the root reason for this issue? I built the app without changing anything and the downloads page loads fine on the first render.

mbroz2:
it should be strictly for performance reasons; currently, the first user has to incur the cost of populating the cache, so it's slower than any subsequent request that would leverage the populated cache.  If instead we populate the cache during server build/start, than that would result in no penalty for the first user.

steven1046:
We can add code to the BuildsManager constructor to initialize the data that dheBuildParser stores, I haven't had time to add a JUnit test to BuildsManagerTest.java. That's as far as I got with this issue. Building the application locally after adding the code makes the downloads page load quick, but I wasn't able to fully test it yet.

deboer-tim:
+1 to this. I happened to be a 'first user'. The placeholder image didn't really make me think that anything was happening in the background, and it took almost a minute before the content appeared. I had time to look through all of the tabs and ping Alasdair about not having any releases before it actually showed up :), so anything to speed it up would be appreciated.

barecode:
The core issue that causes this is that we crawl DHE, because DHE does not have a simple API that informs us in one query what builds are available. That crawling can take a lot of time, so generating this list during deployment / build would be ideal.

NottyCode:
So I've done a PR in #1018 which causes us to load at server start. This isn't quite what this issue is calling for, but doing this at build time is complicated and @mbroz2 was concerned that doing it in build cloud cause issues that would be hard to detect (like if we broke the live function we wouldn't notice).

NottyCode:
In PR #1604 we update how the build population works so during server start the latest release build will be fetched, this means that you won't get in the situation where there are no builds. 

[...]

PR #1589, opened 2020-03-30
1018 force init of the build data on startup rather than 1st request

NottyCode:
Fixes issue #1018

Currently build data is loaded on first request, this can be an issue because the first person to request builds gets nothing. Instead propose a change to run the update when the servletcontainerinitializer is called. It runs inline and on my machine it takes around 6s. This means it will be inited during deployment and not first request.

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/website-api.yml
+++ b/website-api.yml
@@ -1,0 +1,115 @@
+openapi: '3.0.2'
+info:
+  title: Open Liberty Website API
+  version: '1.0'
+servers:
+  - url: https://openliberty.io/api/
+components:
+  schemas:
+    update: 
+      type: object
+      properties:
+        last_update_attempt:
+          type: string
+        last_successful_update:
+          type: string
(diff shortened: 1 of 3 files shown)


website-api.yml
```diff
@@ -0,0 +1,115 @@
...
+          type: array
+          items:
+            type: string
+          description: "This is an array of strings whose values are name=value, this should totally be an object"
+paths:
+  /builds:
+    get:
+      responses:
+        '200':
+          description: "The timestamp for the laste attempt to update the builds, and the last successful run"
```
mbroz2: "laste" -> "last"

PR #1593, opened 2020-04-01
Update codebase to use Jakarta EE 8

NottyCode:
This commit makes the following changes:

1. Update to use Java SE 8
2. Update to build against Jakarta EE 8 APIs
3. Update to use JAX-RS 2.1, CDI 2.0, JSON-P 1.1
4. Use concurrent-1.0 for threads rather than a Java SE fork join pool
5. Use the most recent Liberty maven plugin
6. Remove the ContextListener which is doing a blocking update of the
build data which appears to cause issues.

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


src/main/java/io/openliberty/website/dheclient/DHEBuildParser.java
```diff
@@ -149,6 +152,13 @@ private synchronized void clearScheduledUpdate() {
 		scheduledUpdate = null;
 	}
 
+	private <T> Future<T> submit(Callable<T> c) {
+//		if (exec == null) {
```
steven1046, line 156: Will this be potentially used later?

NottyCode: no. I had at one point had that for test, but it was causing it to be harder to debug in the server so I ended up moving it to the test constructor and didn't remove this code. Fixed now though.

PR #1604, opened 2020-04-18
Rewrite of build management

NottyCode:
The purpose of this PR is to simplify and improve the build management
making better use of Jakarta EE 8 and MicroProfile APIs. In short this
makes the following changes:

1. Make use of MicroProfile REST client for talking to DHE rather than
the Jakarta EE dynamic rest client (simpler client code)
2. Make use of JSON-B rather than JSON-P (also simpler client code)
3. Use a ManagedScheduledExecutorService for async code rather than
making use of unmanaged Java SE threads
4. Simplfiy the thread management to use a fixed rate schedule rather
than calling submit on an executor service after the work is done.
5. Support lazy fetching of data rather than requiring all data to
arrive.
6. Make sure the latest releases are fetched up front during website
initialization
7. Switch to the Java Time APIs

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


NottyCode:
This fix will make #1018 go away since it causes us to always have at least one tools and runtime build prior to the web application being available.

PR #1657, opened 2020-05-14
Add jekyll-include-cache plugin to improve jekyll build time by 10%

steven1046:
This plugin caches certain liquid includes so liquid doesn't have to process the same html multiple times during the jekyll build. Common files included like the header and footer are loaded so many times that caching it makes sense if they are not page specific.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/scripts/build_gem_dependencies.sh
+++ b/scripts/build_gem_dependencies.sh
@@ -1,5 +1,5 @@
 echo "Installing ruby packages..."
 gem install jekyll -v 3.8.6
-gem install bundler jekyll-feed jekyll-asciidoc coderay uglifier octopress-minify-html octokit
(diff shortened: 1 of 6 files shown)


steven1046:
Normal jekyll build:
image

Build using the cache plugin:
image

steven1046:
After testing further, i don't notice any significant difference. I think its because the cached files are so small that it doesn't save a lot of time.

steven1046:
We can see if the build time improves much once the consistent nav gets delivered and also once there's no breadcrumb in javadocs so we can cache the doc-header too.

PR #1714, opened 2020-06-25
Add copy code button to docs, blogs, downloads

ellenwyllie:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/css/config.scss
+++ b/src/main/content/_assets/css/config.scss
@@ -32,6 +32,10 @@
     }
 }
 
+pre {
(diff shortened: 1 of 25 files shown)


steven1046:
We should make it a common function to copy things to the clipboard, might be too hard at the moment since each page has a different tag to copy but after antora we should address this since a lot of iframes will go away.

PR #1774, opened 2020-08-04
Change the server configuration landing page redirect; Add redirects …

steven1046:
…from old server config and feature pages to new ones

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/webapp/WEB-INF/redirects.properties
+++ b/src/main/webapp/WEB-INF/redirects.properties
@@ -30,6 +30,12 @@
 /docs/ref/javaee/=/docs/ref/javaee/8/
 /docs/ref/microprofile/=/docs/ref/microprofile/3.3/
 /javadocs/*=/docs/ref/javadocs/
+/docs/ref/feature/featureOverview.html=/docs/latest/feature/feature-overview.html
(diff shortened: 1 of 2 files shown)


src/main/webapp/WEB-INF/redirects.properties
```diff
@@ -30,6 +30,12 @@
 /docs/ref/javaee/=/docs/ref/javaee/8/
 /docs/ref/microprofile/=/docs/ref/microprofile/3.3/
 /javadocs/*=/docs/ref/javadocs/
+/docs/ref/feature/featureOverview.html=/docs/latest/feature/feature-overview.html
```
mbroz2, line 33: this is fine for now, but we need to open an issue to change this such that we don't have to update this every release.  

So adding support for something like this:
/docs/(.*)$/reference/feature/featureOverview.html=/docs/$1/reference/feature/feature-overview.html

In other words, I want to add support for regex in our redirects, or add variable support to our existing one (performance will need to be considered).

PR #1780, opened 2020-08-06
Change the redirectFilter.java to handle pre-Antora doc urls

steven1046:
Change the redirect filter to not redirect uris that end with the rule if they are a wildcard redirect;
Clean up the doc redirects and add comments;
Add server side redirects from pre-Antora urls to post-Antora latest urls

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/webapp/WEB-INF/redirects.properties
+++ b/src/main/webapp/WEB-INF/redirects.properties
@@ -12,30 +12,51 @@
 /community/=/support/
 /community=/support/
 
-# Blogs
+# Blogs redirects. NOTE: These will move to the blogs repo in the future for customization by the blogs team.
+/news/=/blog/
 /news/*=/blog/
 /blog/2017/11/29/liberty-spring-boot.html=/guides/spring-boot.html
 /blog/2019/03/01/microprofile-concurrency.html=/blog/2019/08/16/microprofile-context-propagation.html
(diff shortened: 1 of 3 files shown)


src/main/webapp/WEB-INF/redirects.properties
```diff
@@ -12,30 +12,51 @@
...
+
+# Redirect 'latest' docs to the latest doc version. Latest is used in the URL by the Doc macros at https://github.com/OpenLiberty/docs/blob/vNext/lib/ol-asciidoc.js. Latest is also used in links to always link to the most recent doc version even when new versions are released.
 /docs/latest/*=/docs/20.0.0.8/
+/docs/latest/=/docs/20.0.0.8/
+
+# Redirect the pre-Antora docs URLS to the new URL format post-Antora
+/docs/ref/general/*=/docs/latest/reference/
+/docs/ref/config/*=/docs/latest/reference/config/
+/docs/ref/command/*=/docs/latest/reference/command/
+/docs/ref/feature/*=/docs/latest/reference/feature/
```
mbroz2, line 49: Lets add another comment after these along the lines of:
"Note that we are purposefully not redirecting the above URLs if they're just ending with "/" (aka nothing for the wildcard to match) because that's handled client-side in order to redirect our old hash based URLs via JS."

PR #1801, opened 2020-08-13
Add Support for Open Graph for ol.io

mbroz2:
Adds support for Open Graph
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ X] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_includes/head.html
+++ b/src/main/content/_includes/head.html
@@ -47,6 +47,11 @@
   <meta name="twitter:title" content="{% if page.seo-title %}{{ page.seo-title }}{% else %}{{site.title}}{% endif %}" />
   <meta name="twitter:description" content="{{ description }}" />
   <meta name="twitter:image" content="https://openliberty.io/img/twitter_card.jpg" />
+
+  <meta property='og:title' content="{% if page.seo-title %}{{ page.seo-title }}{% else %}{{site.title}}{% endif %}" />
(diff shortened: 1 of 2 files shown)


steven1046:
We probably want to add the same meta tags in the doc build in head-meta.hbs.

src/main/content/_includes/head.html
```diff
@@ -47,6 +47,11 @@
   <meta name="twitter:title" content="{% if page.seo-title %}{{ page.seo-title }}{% else %}{{site.title}}{% endif %}" />
   <meta name="twitter:description" content="{{ description }}" />
   <meta name="twitter:image" content="https://openliberty.io/img/twitter_card.jpg" />
+
+  <meta property='og:title' content="{% if page.seo-title %}{{ page.seo-title }}{% else %}{{site.title}}{% endif %}" />
+  <meta property='og:image' content="{% if page.open-graph-image %} {{page.open-graph-image}} {% else %}https://openliberty.io/img/twitter_card.jpg{% endif %}" />
+  <meta property='og:description' content="{{ description }}" />
+  <meta property='og:url' content="{{ page.url | replace:'index.html','' | absolute_url }}" />
```
ellenwyllie, line 54: The opengraph website says 'type' is a required property so we might need to add that as well.

mbroz2: yes, not sure how I dropped that from the commit, will add it back along with the update to head-meta for docs

mbroz2:
Example docs output after adding OG:
image

PR #1825, opened 2020-08-24
Update sitemap for docs

mbroz2:
Also removes lastmod for each url as we were not updating them (if we move to using automation to create the sitemap, we can have it add it back in).

### DIFF
--- a/src/main/content/sitemap.xml
+++ b/src/main/content/sitemap.xml
@@ -5,37 +5,30 @@
 
     <url>
         <loc>https://openliberty.io/</loc>
-        <lastmod>2020-02-10T21:04:44+00:00</lastmod>
-        <changefreq>weekly</changefreq>
+        <changefreq>monthly</changefreq>
     </url>
     <url>
-        <loc>https://openliberty.io/docs/</loc>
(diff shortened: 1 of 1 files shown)


src/main/content/sitemap.xml
```diff
@@ -5,37 +5,30 @@
...
         <loc>https://openliberty.io/</loc>
-        <lastmod>2020-02-10T21:04:44+00:00</lastmod>
-        <changefreq>weekly</changefreq>
+        <changefreq>monthly</changefreq>
     </url>
     <url>
-        <loc>https://openliberty.io/docs/</loc>
-        <lastmod>2019-02-12T21:04:44+00:00</lastmod>
-        <changefreq>weekly</changefreq>
+        <loc>https://openliberty.io/docs/20.0.0.8/overview.html</loc>
```
steven1046, line 11: Can we just specify https://openliberty.io/docs/ for this which gets redirected to the right location or do we have to change it every release?

mbroz2: unfortunately sitemaps are not suppose to have redirects

Issue #1820, opened 2020-08-20
Write an extension for Antora's version sorter


steven1046:
Right now we use `X.0.0.X` as our version schema, and Antora only seems to support `X.X.X` following the convention https://semver.org/. This messes up our canonical urls and version picker sort order because Antora doesn't recognize our version ordering with the third period `.` in the version. We need to write an extension to the tooling to handle sorting versions with this many version components.

steven1046:
We wrote a custom helper (#1849) to sort our versions so they are in order. In the future, Antora might expose a version extension here: https://gitlab.com/antora/antora/-/issues/454.

steven1046:
Look up how to extend the node_module that Antora pulls in called "semver".

k8vance88:
Closing this issue as it's already working.  I see no reason to change it.

PR #1845, opened 2020-09-02
Reverse the versions array so the highest version is on top

steven1046:
This will reverse the versions of our docs in the version picker so the highest version is on top.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
This may or may not make #1820 irrelevant to do.

PR #1928, opened 2020-11-05
Move javadoc repo clone to after jekyll command

steven1046:
This is so jekyll doesn't process the javadocs
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/scripts/build_jekyll_maven.sh
+++ b/scripts/build_jekyll_maven.sh
@@ -45,22 +45,6 @@ if [ "$JEKYLL_ENV" != "production" ]; then
     fi
 fi
 
-# Determine which branch of docs-javadoc repo to clone
-BRANCH_NAME="prod"
-if [ "$JEKYLL_STAGING_SITE" == "true" ]; then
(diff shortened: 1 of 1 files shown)


mbroz2 (approved):

mbroz2:
Did you happen to notice a build performance improvement with this?

steven1046:
> Did you happen to notice a build performance improvement with this?

It's hard to notice a huge improvement using Travis. Should i do some testing or just merge it in?

mbroz2:
No need, I can't see how it could cause a degradation in performance, was just hopeful you might have noticed some meaningful performance improvement.

steven1046:
@mbroz2 I've seen some performance improvements from other changes in the past using multiple builds and averages, but this one i'm unsure if there would be much of a difference or if it's even worth testing because it should be the same time or less for builds after this.

PR #2037, opened 2021-02-15
Fix accessibility issue

dmorgant:
Accessibility Checker flagged an issue with the TOC <div> element.

image

This fixes the issue by setting the TOC `<div>` that is visible when the window width is thin so that the guide appears as a single column to have `role="button"`.   The `<div>` acts as a button causing the TOC to appear when it is selected and 'aria-expanded' is valid for a button.

- [x] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
- [ ] 
#### Running validation tools
- [x] Accessibility Checker

This issue was fixed as part of https://github.com/openliberty/iguide-microprofile-config-intro/issues/254

### DIFF
--- a/src/main/content/_layouts/guide-multipane.html
+++ b/src/main/content/_layouts/guide-multipane.html
@@ -28,7 +28,7 @@ js:
 <!-- Remove any documents deemed private (archived, templates, etc) -->
 {% for guide in all-markdown-files %}
     {% unless guide.archived or guide.path contains "guides/guides-template" or guide.path contains "guides/guides-common" or guide.path contains "guides/iguides-common" %}
-        {% assign all-guides = all-guides | push: guide %} 
+        {% assign all-guides = all-guides | push: guide %}
     {% endunless %}
 {% endfor %}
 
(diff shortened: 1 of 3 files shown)


steven1046:
While we found the fix to this issue, could we also change the div role for static guides in https://github.com/OpenLiberty/openliberty.io/blob/prod/src/main/content/_layouts/guide.html and https://github.com/OpenLiberty/openliberty.io/blob/prod/src/main/content/_layouts/guide-multipane.html to use the same role?

dmorgant:
Sorry! ....had no idea we had all that duplicated code.    Fixed in those two files also.

src/main/content/_layouts/guide-multipane.html
```diff
@@ -112,7 +112,7 @@ <h1 id="guide_title">{{ page.title }}</h1>
                 </div>
 
                 <div class="scroller_anchor"></div>
-                <div role="menu" id="mobile_toc_accordion_container" data-toggle="collapse" data-target="#toc_column" aria-expanded="false" aria-controls="toc_column">
```
steven1046, line 115: Can you remove the old role="menu" from these please

src/main/content/_layouts/guide.html
```diff
@@ -109,7 +109,7 @@ <h1 id="guide_title">{{ page.title }}</h1>
                 </div>
 
                 <div class="scroller_anchor"></div>
-                <div role="menu" id="mobile_toc_accordion_container" data-toggle="collapse" data-target="#toc_column" aria-expanded="false" aria-controls="toc_column">
```
steven1046, line 112: Can you remove the old role="menu" from these please

steven1046 (approved):

Issue #1924, opened 2020-11-03
At certain small browser widths, the TOC indicator arrow doesn't work


steven1046:
![image](https://user-images.githubusercontent.com/6392944/98015972-04208380-1dc3-11eb-82f0-6c7b733b1453.png)
If you click the orange arrow, you expect the TOC to open up but at certain widths in 2 column view it animates to the same section you're on and doesn't open the TOC.

steven1046:
So it does open the TOC, but the TOC is stickied to the top of the guide and you can't see it from where you're at.

steven1046:
It might be a gap in the css @media ranges

dmorgant:
For me, I am having problems with the TOC in a different way.    When I scroll down pass the nav header, I see the following, where the guide content is shifted to the left and a blank column, about the width of the TOC, is in the middle of the page.

![image](https://user-images.githubusercontent.com/29490139/109863327-20281a80-7c27-11eb-99d1-ab34c6ae71bc.png)

Then, when you get to the end of the guide the TOC displays again, although looks transparent as it is written on top of the guide end.

![image](https://user-images.githubusercontent.com/29490139/109863229-fff85b80-7c26-11eb-9293-a75f2a6fc27a.png)

---------------------------------------

I also see a problem that when scrolled all the way to the top of the guide the TOC header (where it says 'Contents' and the 'x' to close the TOC) is not viewable.  It is hidden under the nav bar.

![image](https://user-images.githubusercontent.com/29490139/109863475-4b126e80-7c27-11eb-8af5-cee1e4b838d0.png)

It should look like this

![image](https://user-images.githubusercontent.com/29490139/109863513-56fe3080-7c27-11eb-881e-16cbd4736a92.png)

dmorgant:
This may also fix the problem described in #1878

Issue #1878, opened 2020-09-21
Guide Pages Display Error


scottcurtis2605:
On any of the guides at https://openliberty.io/guides/, the menu on the left side of the screen displays incorrectly on both Firefox (vers. 80.0.1) and Safari (vers. 13.1.2) on 13" MacBook Pro 2019 machines (macOS Catalina vers. 10.15.6). The menu bar, within the `<div>` element with `id=background_container` hangs over the `<div>` element with `id=end_of_guide`, which makes the items within the `end_of_guide` container difficult to access. I suspect this is due to the resolution of the Mac - 2560x1600. See the attached image for reference. 

<img width="1438" alt="Screenshot 2020-09-21 at 13 52 38" src="https://user-images.githubusercontent.com/41568419/93769969-7ffbad80-fc13-11ea-8b9a-c5bde31e8ac8.png">

This issue did not persist on a Windows 10 (vers. 1903) machine in my possession, with a resolution of 1920x1080 on both Firefox (vers. 80.0.1), and Chrome (vers. 85.0.x).

steven1046:
Hi @scottcurtis2605, I am unable to produce this bug on Safari or Firefox (on Mac). Do you have any tips on how you saw this?

steven1046:
Nevermind, i was able to reproduce while zoomed in far enough and at the bottom of the guide, the table of contents is overlaid wrong.

dmorgant:
I believe this is fixed with issue #1924.   Will review again once the fix makes it to prod.

dmorgant:
With the fix for issue #1924 Steven and I are no longer able to recreate this problem.   

Closing this issue.  If it still occurs, please reopen with details.  Thank you.

PR #2054, opened 2021-03-11
Merge in TOC formatting change from Staging

dmorgant:
Copied from #2051.

Fixes issue #1924 and #1878.

I was unable to recreate the problem that Steven described in the issue. My problems with the TOC displayed improperly are also described in the issue. I traced it down to having a zoom level in my browser window other than 100%. This was causing window.outerWidth to return what seemed to be an incorrect value to designate one-, two-, or three-column formatting for the guide. It turned out that window.innerWidth seemed to be the value we were after in toc.js to determine the layout of the guide, and in looking at the code, had been used in many of our calculations already. In particular, the change from window.outerWidth to window.innerWidth made in toc.js on line 18 solved the majority of both problems Steven and I were seeing. I also made several other changes from outWidth to innerWidth and tested each one and they seem to solve the other problems described in issue #1924 as well.

It should be noted that we use window.outerWidth in guides.js, which executes when the user is on the openliberty.io/guides page. But switching out outerWidth to innerWidth in this file caused issues, especially going from one- to two-columns in the smaller widths. So, since I did not see any obvious formatting issues on this page I decided to leave guides.js alone.

- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/js/openliberty.js
+++ b/src/main/content/_assets/js/openliberty.js
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 IBM Corporation and others.
+ * Copyright (c) 2019, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


dmorgant:
This may also fix the issue described in #1878.

steven1046 (approved):

PR #2152, opened 2021-06-11
Parse blogs json during build time and add it to blog front matter

### LINKED ISSUE #1934: Jakarta EE feed doesn't pull in all of the "Jakarta EE" tagged blogs
This is because we externalized our blog tags to a single file in the blogs repo (https://github.com/OpenLiberty/blogs). Jekyll doesn't know about the tags from each blog anymore since it uses site.tags to process the feed and site.tags is emptied now that blogs don't have tags in each file.

https://openliberty.io/jakartaee.xml only has one post because one file https://raw.githubusercontent.com/OpenLiberty/blogs/prod/posts/2019-05-08-jakarta-ee-naming.adoc still has a tag in the post by mistake.

steven1046:
Since we removed tags from individual blog posts, a temporary fix was put in to add "JakartaEE" tags to those blog files individually. Since our tags are stored in a central file, https://github.com/OpenLiberty/blogs/blob/prod/blog_tags.json, we should use those instead and read in those tags during build time and place them back in the blogs so Jekyll is aware of the tags when processing things like feeds, and using site.tags.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/scripts/build/jekyll.sh
+++ b/scripts/build/jekyll.sh
@@ -58,6 +58,9 @@ $BUILD_SCRIPTS_DIR/clone_certifications.sh
 # Clone draft and published blogs
 $BUILD_SCRIPTS_DIR/clone_blogs.sh
 
+# Read in the blog tags file and add the tag to each blog so jekyll knows how to process them.
+python3 ./script/build/parse_blog_tags.py
+
 # Jekyll build all the cloned content
 echo "Building with jekyll..."
 echo `jekyll -version`
--- a/scripts/build/parse_blog_tags.py
+++ b/scripts/build/parse_blog_tags.py
@@ -7,9 +7,9 @@ tags = content['blog_tags']
 for tag in tags:
     tag_name = tag['name']
     posts = tag['posts']
-    for post in posts:
+    for post_name in posts:
         for file_name in os.listdir("src/main/content/_posts/"):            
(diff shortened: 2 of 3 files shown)


steven1046:
~This needs https://github.com/OpenLiberty/blogs/pull/1485 to work, as the tags: front matter doesn't need a bracket for one tag and the script expects the tags front matter to just be a list of tags with space delimiters.~

This work no longer needs the above pr to work, but it will be nice to remove those tags because they won't be used anymore. The blog json file uses "Jakarta EE" tag, and our jakarta.xml feed looks for that tag with a space now, so we no longer need the "JakartaEE" tag in those 4-5 posts.

steven1046:
I still need to add brackets around the tags in the front-matter, because our tags have spaces and jekyll will separate those tags by space.

steven1046:
I verified that http://localhost:4000/jakartaee.xml still has all of the Jakarta EE posts after removing the tags from the blogs repo. It actually has more, all of the https://openliberty.io/blog/?search=Jakarta_EE&key=tag posts.

natalie-bernhard (approved):

PR #2191, opened 2021-06-30
Fix hotspot contains bug

### LINKED ISSUE #1650: Allow guide hotspots to look for exact instead of partial match
When writing guides and using hotspots, I noticed that the hotspot would also match start and end tags partially. For example, I have a hotspotted `deploy.yaml` file:
```
# tag::system[]
apiVersion: openliberty.io/v1beta1
kind: OpenLibertyApplication
metadata:
  name: system
  labels:
    name: system
spec:
  # tag::systemImage[]
  applicationImage: guide/system-imagestream:1.0-SNAPSHOT
  # end::systemImage[]
  pullPolicy: Always
  env:
    - name: MP_MESSAGING_CONNECTOR_LIBERTY_KAFKA_BOOTSTRAP_SERVERS
      value: "kafka:9092"
# end::system[]
```

I added a hotspot in the guide that references the whole `system` tag block: `the [hotspot=system file=1]'system' and ...` which I expected to highlight the whole block, however, on the website, it seems that the `# end::systemImage[]` tag is where the highlight ends:

image

Could the component that highlights the codepane search for an exact match instead of partial to avoid situations like this? If this is too tedious let us know, we can apply a workaround to our hotspot naming convention to avoid edge cases like this.

Thanks!

steven1046:
Change the end of a hotspot lines to highlight; to end with [] so it doesn't match substrings of other hotspots
Fixes #1650 
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/js/guide-multipane-static.js
+++ b/src/main/content/_assets/js/guide-multipane-static.js
@@ -18,18 +18,28 @@ var twoColumnBreakpoint = 900; // This overrides the twoColumnBreakpoint set in 
 // Input: Code_block: the code file
 //        Header: the section header for this code file
 //        Index: the index of the file in this section
-function link_hotspots_to_file(code_block, header, index){
+function link_hotspots_to_file(code_block, header, index) {
     // Check how many code_column are present in this subsection.
-    var sect = code_block.parents('.sect1');
-    var num_files = $(header).siblings('.code_column').length;   
+    var sect = code_block.parents(".sect1");
+    var num_files = $(header).siblings(".code_column").length;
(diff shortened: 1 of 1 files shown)


steven1046:
@natalie-bernhard the only actual non-formatting change was this:
var end = $(this)
            .nextAll("span:contains('end::" + tag_name + "[]')")
            .first();

natalie-bernhard (approved):

PR #2228, opened 2021-07-28
Gzip doc html

### LINKED ISSUE #1623: Enable HTTP Compression (gzip)
As of 20.0.0.4, support for http compression has been added:
https://openliberty.io/blog/2020/04/22/http-response-compression.html

In order to enable this, we'd need to add the necessary compression element to httpEndpoint.  However, we currently do not specify our server.xml when pushing using cf (we push a stand-alone app), so we'd need to switch to doing something like using the `-p` flag and provide a server directory, which would allow us to include our own custom server.xml.  Another option would be to push a packaged server, but that would have extra overhead.
For documentation, see here: https://cloud.ibm.com/docs/runtimes/liberty?topic=liberty-options_for_pushing

steven1046:
The app size got too large (>1GB) when pushing to cloud foundry and we need to reduce the size of the app so we can make updates. Gzipping the docs html reduces the size of the app from roughly 1.4GB on disk to ~758 MB.
Added logic to not do any extra filters if we are returning gzipped content in TLSFilter.java. It will continue to the redirect filter if it checks and sees that the URI matches any redirects in the servlet context.
I changed the deploy function to deploy a whole server directory instead of using the default Liberty buildpak. This allows us to modify the server.xml that is used for the website and to add gzip compression. 

Removed a duplicate feature enabled in the manifest.yml.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
This requires https://github.com/OpenLiberty/docs-playbook/pull/102 to be merged to work.

steven1046:
This is to resolve the error when deploying our app:
`Waiting for API to complete processing files...
Job (7686ee66-773c-453d-adcc-aea8067f3065) failed: The app package is invalid: Package may not be larger than 1073741824 bytes`

natalie-bernhard (approved):

PR #2251, opened 2021-08-10
Javadoc link fix

### LINKED ISSUE #316: Make javadoc classes and packages use the same URL schema when copying their link
Currently in the MicroProfile 1.3 javadoc if you copy the link of a package you would get https://openliberty.io/javadocs/microprofile-1.3-javadoc/javax/annotation/package-frame.html when the URL is actually https://openliberty.io/docs/ref/microprofile/1.3/#package=javax/annotation/package-frame.html&class=overview-summary.html so there is an inconsistency. If the user goes to the former URL then they would not see the Open Liberty header.

steven1046:
I modified the feature script to have the hrefs be the shorthand path like they were before while comparing them to regex, and then switch the href back to the full path afterwards so it matches the rest of the doc hrefs.

Renamed the feature script to make more sense and not be so different than other script names.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
Marking the pr as draft until Java EE has been addressed too

Issue #2218, opened 2021-07-13
Prevent crawling noversion.adoc in docs


steven1046:
Searching Google for open liberty 20.0.0.7 docs shows some people "Document does not exist" as a result which may confuse people. We should look to prevent crawling of https://github.com/OpenLiberty/docs/blob/vNext/modules/ROOT/pages/noversion.adoc because that is only meant to be shown when a doc doesn't exist when using the version switcher.

steven1046:
noindex.hbs:
`<meta name="robots" content="noindex" />`

needs to be added to this file

steven1046:
We have this check if it's a non prod site for jekyll:
```  
{% if jekyll.environment != 'production' or page.permalink == '/blog/redirected.html' %}
      {% include_cached noindex.html %}
  {% endif %}
```
  
If that is true, then noindex.html gets built by jekyll , and then later in the build process we copy over noindex.html into Antora's head files in noindex.hbs (handlebars).
  
We need something similar to be done in the Antora head-meta.hbs file. We need to change this file: https://github.com/OpenLiberty/openliberty.io/blob/prod/scripts/build/antora_install.sh#L19 to always copy it into noindex.hbs and then check in head-meta.hbs whether to include it in the head tag of the page depending if it should be indexed or not (include it for any non-prod site and also include it if it's production AND it's noversion.adoc which is in the doc repo).

PR #2258, opened 2021-08-13
Noversion crawling

natalie-bernhard:
Issue #2218 
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
I think adding an if statement around the noindex would disable it for all other doc files, so we might have to find another way to add it to noversion

steven1046:
Right now the build copies over the `<meta name="robots" content="noindex" />` into noindex.hbs if the site isn't production, in antora_install.sh.

PR #2269, opened 2021-09-08
Fix bug in calculating the highest doc version

steven1046:
This pr fixes a bug in python where max(versions) does not compute the maximum doc version. It was computing 21.0.0.9 over 21.0.0.10 so therefore it wasn't computing each part of the version as an integer. This should fix the computation, so the draft site will have the correct features combined.
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
@mbroz2 @k8vance88 i tested it locally.

mbroz2 (approved):

scripts/build/parse_features_toc.py
```diff
@@ -46,12 +46,26 @@ def createVersionHref(parent, tocHref, tocString):
...
 # Only process the features of the highest version for the docs draft site
 if(os.getenv("DRAFT_SITE") or os.getenv("DOCS_DRAFT_SITE")):
-    print("Processing only version: " + max(versions) + " for the docs draft site.")
-    versions = [max(versions)]
+    max = ['0', '0', '0', '0']
+    for version in versions:
+        nums = version.split('.')
+        if(int(nums[0]) > int(max[0])):
+            max = nums
+        elif(int(nums[0]) == int(max[0]) and int(nums[3]) > int(max[3])):
```
k8vance88, line 59: This works as long as we always stick to zeros in the middle.

k8vance88 (approved):

scripts/build/parse_features_toc.py
steven1046: As far as i know, we'll always have the zeros.

PR #2281, opened 2021-09-27
Get started

mbroz2:
image

Also confirmed the following redirects work:
/downloads
/downloads/


steven1046:
Should we also update most links going to /downloads to be /start such as in the header.html?

index.html
sitemap.xml
header.html
header-content.hbs

mbroz2:
It wouldn't hurt, but I think I'd like us to do that under a separate 'cleanup /downloads links' issue rather then this one.

steven1046 (approved):

mbroz2:
https://github.com/OpenLiberty/openliberty.io/issues/2282

PR #2318, opened 2021-10-25
Support doc feature pages being dropdowns and an actual page

steven1046:
This is to fix the issue in https://draft-openlibertyio.mybluemix.net/docs/21.0.0.11/reference/feature/feature-overview.html where the features aren't getting combined because previously the python script was looking to combine features underneath the `span` with the word 'Features' and now this will support both the span for older doc version TOC's as well as the new `anchor` dropdowns where the Feature overview dropdown is a clickable page itself.
Fixes this issue with the features not getting combined:
image

Also fix scrolling to the right spot in the toc when features load.
Also change the styling of the nav so these dropdown pages are on the same line as their arrows:
image

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/scripts/build/parse_features_toc.py
+++ b/scripts/build/parse_features_toc.py
@@ -79,119 +79,128 @@ for version in versions:
     toc = featureIndex.find('nav', {'class': 'nav-menu'})
     if toc.find('span', text='Features') is not None:
         featureDropdown = toc.find('span', text='Features').parent
-        for featureTOC in featureDropdown.find_all('a', {'class': 'nav-link'}, href=True):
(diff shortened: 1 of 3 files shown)


steven1046:
Updates can be seen at https://draft-openlibertyio.mybluemix.net/docs/21.0.0.11/reference/feature/appSecurity-3.0.html. Please merge if approved.

mbroz2 (approved):

Issue #2377, opened 2021-11-30
Allowlist approach for displaying zips from DHE on /start


mbroz2:
Our code that grabs the zips from DHE to display on /start is too greedy.  It should instead only expose the archives that we expect and not everything that it finds in the folder.

For example, today, we show the MP5 beta (https://openliberty.io/start/#runtime_betas) in the betas tab, but call it 'All Beta Features'.  Instead, we should have had only a single entry for All Beta Features pointing at https://public.dhe.ibm.com/ibmdl/export/pub/software/openliberty/runtime/beta/2021-11-17_1256/openliberty-22.0.0.1-beta.zip and ignored the unexpected MP5 beta.

kinueng:
Closed as fixed
https://openliberty.io/start/#runtime_betas
![image](https://user-images.githubusercontent.com/31117513/145278228-1eb85ce1-71b6-4bec-9d35-04a7dff2d2a6.png)

kinueng:
We need a cleaner fix that applies to all of the tabs.  Reopening to attach more PRs to this issue

kinueng:
There are some conversations within PR #2406 that may interest anyone looking for history on the v1.0 allowed list logic implementation

PR #2405, opened 2021-12-09
Use an allow list for determining which builds show on the website (draft)

kinueng:

For #2377 

- Move away from blindly showing all zips found on DHE. Use a hardcoded
list of zips that we expect to be listed on the website.

- It is tricky to construct the allowed list for the Betas zips because
beta zips are labeled by the liberty version. This is handled by a
function to construct the allow list on the fly.

- There can be more than one zip per beta version. Replace the previous
fix from c0f13d4b55a62fb55833aaa0a884966ff04ae78a with the new allow
list logic.

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


kinueng:
Since this is just to proof on draft, merging.

PR #2406, opened 2021-12-09
Use an allow list for determining which builds show on the website (staging)

kinueng:

For #2377 

- Move away from blindly showing all zips found on DHE. Use a hardcoded
list of zips that we expect to be listed on the website.

- It is tricky to construct the allowed list for the Betas zips because
beta zips are labeled by the liberty version. This is handled by a
function to construct the allow list on the fly.

- There can be more than one zip per beta version. Replace the previous
fix from c0f13d4b55a62fb55833aaa0a884966ff04ae78a with the new allow
list logic.

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_assets/js/builds.js
+++ b/src/main/content/_assets/js/builds.js
@@ -24,6 +24,70 @@ var builds_url = '/api/builds/data';
 var starter_info_url = 'https://start.openliberty.io/api/start/info';
 var starter_submit_url = 'https://start.openliberty.io/api/start';
 
+// Controls what build zips are exposed on openliberty.io.  This will need to be updated
+// if there is a new zip version published on DHE.  The intent of this allow_builds list is to
+// prevent the situation where unintential zips on DHE get shown on the website.
+var allowed_builds = {
+    // allow_builds list based on keys found in `package_locations` 
+    // from the https://openliberty.io/api/builds/data.  The keys (i.e. kernel.zip) are defined by
+    // by BuildInfo.java, method resolveLocations(String url, BuildType type, String dateTime)
+    //
+    // For example:
+    // {
+    //     "builds": {
+    //         "runtime_releases": [
+    //             {
+    //                 "package_locations": [
+    //                     "javaee8.zip=...openliberty-javaee8-21.0.0.12.zip",
+    //                     "jakartaee9.zip=...openliberty-jakartaee9-21.0.0.12.zip",
+    //                     "webProfile8.zip=...openliberty-webProfile8-21.0.0.12.zip",
+    //                     "webProfile9.zip=...openliberty-webProfile9-21.0.0.12.zip",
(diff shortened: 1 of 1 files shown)


kinueng:
I used Beyond Compare to compared the HTML between https://openliberty.io/start/  and https://ui-draft-openlibertyio.mybluemix.net/start/.  There was no difference in visible builds.

mbroz2 (approved):
This works.

Since we already process the DHE data on the server backend and present the manipulated results via our own API, I probably would have implemented this on the server side to avoid processing this each time on the client.  I would have also moved the hardcoded labels/names for these packages to the API/server-side as well so that the client side just displayed the contents of API payload without needing to 'process' the data.

However, given that the processing of this should be cheap/quick, I think it's fine to keep it as-is.

kinueng:
I went back and forth in my head whether I should change the Java code for `https://openliberty.io/api/builds/data`.

I ultimately decided to process the allowed list on the JavaScript side. My thoughts are that we should preserve the idea that `https://openliberty.io/api/builds/data` is just a wrapper (with some extra caching/performance logic) for the already publicly accessible data on DHE.  The API is not in the role of curating what is found at `https://public.dhe.ibm.com/ibmdl/export/pub/software/openliberty/runtime/`.  If there is something that shouldn't be on DHE, it should be removed from DHE rather than at our website API.

mbroz2:
Unlike the builds/data API, openliberty.io is not the only consumer of content hosted on DHE.  That are valid reasons for why we might have content on DHE that should not be exposed on the website, which is at least partially the reason for why we're adding the allow-list in the first place.

I think your question regarding the intent/purpose of the /api/builds/data is a good one; if it's meant to just be a cache of DHE content then I agree that doing the processing on the client might make more sense.  However, my view of the API is that it's not just a cache of DHE, but rather an API that provides the necessary data for the builds tables (otherwise I would have expected it to be /api/dhe/cache/ not /api/builds/data/).

Keep in mind that we have intent to add even more content to the tables (checksums and signatures), which may add even more client side processing to the page.  However, as long as these changes don't affect the load time for the page for users, then it's a non-issue where the data is processed.

kinueng:
Oh, so the `/api/builds/data` endpoint does curate data from DHE already, which goes against the idea that the endpoint is just a wrapper.

The `.md5` and `.sha2` files are not exposed by the endpoint, but are on DHE.  Checksums file examples [here](https://public.dhe.ibm.com/ibmdl/export/pub/software/openliberty/runtime/nightly/2021-12-10_0301/) on DHE

kinueng:
A better way of describing the website endpoint is that it is a wrapper for the `info.json` files found on DHE.

PR #2436, opened 2022-01-18
Javadoc query params

### LINKED ISSUE #2301: Change javadoc urls from using `#` to `?`
Currently our javadoc URLs use hashes, which works, but it means that all the javadocs will not be indexed by search engines.  Instead, we should change these to use query params and redirect the hashes to it.

Since we currently don't have sitewide search, this is a problem since a user would need to manually find what they're looking for in a javadoc instead of being able to use a search engine.

steven1046:
Change javadocs from storing the package location and class location in the url from hash to query parameters:
Example:
#package=javax/annotation/package-frame.html&class=javax/annotation/Generated.html
changes to:
?class=javax/annotation/Generated.html&package=allclasses-frame.html

I changed the javascript we have to redirect old pre-Antora urls to the post-Antora urls so it uses query parameters.
I also changed the javadoc search javascript (that we override the default javadoc search js with) to use query parameters
.
- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
Closes #2301

src/main/content/_assets/js/javadoc-search.js
```diff
@@ -358,26 +360,26 @@ $(function() {
...
                     } else {
-                        hash = "#class=" + url + "&package=allclasses-frame.html"; // putting in default html for package iframe
-                    }
-                    if (hash !== top.window.location.hash) {
-                        var state = {};
-                        state["iframe.rightIframe"] = href;
-                        // create a history entry
-                        top.window.history.pushState(state, null, hash);
+                        queryParams.set('class', url);
+                        queryParams.set('package', 'allclasses-frame.html');
```
kinueng: Can you give me some context on what is happening here?  To my untrained eye, it looks like we are calling `queryParams.set('class', url);` regardless of whether `queryParams.has('class')` is `true` or `false`.

src/main/content/antora_ui/src/js/06-javadocs.js
```diff
@@ -364,19 +364,19 @@ function addNavHoverListener() {
...
-  var hashPage = parent.window.location.hash;
-  if (hashPage != "" && hashPage != undefined) {
-    hashPage = hashPage.substring(1); // take out the #
-    var splitHashPage = hashPage.split("&");
-    for (i = 0; i < splitHashPage.length; i++) {
-      var hashString = splitHashPage[i].trim();
-      if (hashString.indexOf(PACKAGE_HASH) === 0) {
-        targetPage.package = hashString.substring(8);
+  var queryParams = window.location.search;  
+  if (queryParams && queryParams != undefined) {
```
kinueng: Lets use `!==` instead to avoid any unexpected type conversion

src/main/content/antora_ui/src/js/06-javadocs.js
```diff
@@ -561,73 +569,50 @@ function setIFrameContent(iframeName, href) {
...
 //
-// The package hash is used to render the content in the left bottom iframe. The class hash
+// The package param is used to render the content in the left bottom iframe. The class param
 // is used to render the content in the right iframe.
-function setHashParams(url, hashName) {
-  var hash = window.location.hash;
+function setQueryParams(url, paramName) {
+  var newURL = new URL(window.location.href);
+  var queryParams = newURL.searchParams;
   if (url !== undefined && url != "") {
```
kinueng: Let's also use `!==` here

src/main/content/_assets/js/javadoc-search.js
steven1046: I was just changing the logic from hash to QP. I do see why we don't need to set it in both cases, but i'll figure it out today and improve it.

steven1046:
Tested on the demo1 site

kinueng (approved):
LGTM

PR #2450, opened 2022-01-20
2446 add a link2 public key

aknguyen7:


kinueng (approved):
LGTM on draft site.  We can proof it again on staging

PR #2476, opened 2022-02-02
2218

aknguyen7:
Issue #2218

To prevent web crawling, we need to add this line to the page `<head> <meta> `data in docs

image

Currently it's only add for non production, need to also add for production if the page is noversion

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


steven1046:
We only want this robots meta tag to show up on the noversion page on production, not all docs.

steven1046:
I see that you can't check the site.url variable in the handlebar code because i think we don't set `baseUrl` in the doc playbook: https://github.com/OpenLiberty/docs-playbook/blob/prod/antora-playbook.yml so we'll either have to use a handlebars helper like you tried before or another method of checking for prod.

kinueng:
Hi @aknguyen7, this PR should target either `staging` or `draft` branch.

PR #2497, opened 2022-02-09
Increate app timeout

mbroz2:
Likely will have no affect since the default max CF timeout is 180, but it's possible IBM Cloud raised that so it's worth checking (doesn't seem to be documented in IBM Docs)


kinueng (approved):
LGTM, merging

Issue #2261, opened 2021-08-26
Add a 'Next steps' or 'What's next' section after generate project in the starter


mbroz2:
Continues on #2260

After the user clicks "Generate Project' on the starter, display a small section after the button that tells the user what to do next depending on what they selected in the starter.  For example:
unzip
run [mvnw/gradlew command, depending on what was chosen in starter]
link to guides

steven1046:
Is a design needed?

k8vance88:
@mdenby can you take a look at this?

mdenby:
Yes, my first thought is a modal...I'll bring this up in the Thursday meeting tomorrow. I just tried it on the site and I remember this--you get a zip file and no 'success' message or instructions (or link to a guide). 

'Your zip file has been beamed to your computer! Unzip it and follow the README instructions or run this command. For more help visit our helpful guides. Button--Guides. Button--Got it! or just the X to close in the top corner.'

Happy to mock this up too.

mdenby:
Here is what the modal could look like. @lauracowen I'm sure the wording could be much better!

<img width="921" alt="Screen Shot 2021-10-14 at 1 07 15 PM" src="https://user-images.githubusercontent.com/26823018/137368424-d3106be5-2395-4baa-9c47-3756b0a21243.png">

mbroz2:
We'll also want to add the actual command to run, and then reference the README.md for more information.

steven1046:
Should we link to the blog post (https://openliberty.io/blog/2021/08/20/open-liberty-starter.html) in the modal as well?

PR #2528, opened 2022-02-21
'What's next' section after generate project in the starter 

Sreejith-Websphere:
- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


src/main/content/_assets/js/builds.js
```diff
@@ -1002,6 +1002,23 @@ $(document).ready(function () {
                         document.body.removeChild(anchor);
                         url.revokeObjectURL(anchor.href);
                     }, 1);
+                    var genProjModal = document.getElementById("generate-project-modal");
+                    genProjModal.style.display = "block";
```
kinueng: We should avoid inline styles to practice the separation of styles (CSS) from the layouts (HTML).  Additionally, inline styles ignore the cascading stylesheets behavior and can add complexity to how styles are applied to HTML elements.

A suggestion is to have the modal be hidden by default through `.generate-project-modal-class`  and apply a "is visible" class on the modal that takes priority over the default hidden styling when it needs to be presented to the user. The is visible class can then be reused for future modals and unify how modals are made visible on the website.  

Here is how vanilla JS Carbon controls a modal https://the-carbon-components.netlify.app/component/modal--default.html

kinueng (changes requested):

src/main/content/_assets/js/builds.js
Sreejith-Websphere: ok here the modal popup is hidden by default. I will change this inline style by adding class for show and hide popup

Sreejith-Websphere: committed changes - removed inline style in js

kinueng (changes requested):
Here are my findings from running `jekyll s` on branch `next-step-generate-project` using my laptop.

- The current modal does not match the styling of the original proposed design.  The design shows the modal has rounded corners, different `X` icon and positioning, and green outline. 
 See the design here https://github.com/OpenLiberty/openliberty.io/issues/2261#issuecomment-943576810.
 
- My opinion is the modal should be vertically and horizontally centered in the browser window.

- The `[ mvnw liberty:run ]` text needs to be styled like a code block.  It is currently plaintext.

-  The modal should be the primary layer.  A minor issue with the floating cow image.
image

Accessibility:
- I am unable to use the tab key to put the focus on the X icon to dismiss the modal.
- I cannot use the Esc key or click outside the modal to dismiss the modal.
- When I put the focus on `Generate Project` and push Enter, the focus should moved to an element in the modal (e.g. `Got it!` button).  The focus currently remains on `Generate Project` button, which is a background element when the modal is visible.
- The modal should trap the focus.  When focus is in the modal, pushing tab key multiple times should not cause the focus to escape the modal and focus elements in the background (e.g. Generate Project button).
- IBM Equal Access Accessibility Checker browser extension is violation on the X icon.
     - `Text contrast of 2.30 with its background is less than the WCAG AA minimum requirements for text of size 30px and weight of 700`
     - image

src/main/content/_assets/js/builds.js
kinueng: Thank you for putting in the extra work to rework the inline styling

Sreejith-Websphere:
Updated the code - fixing issues. @kinueng please review it

src/main/content/_assets/css/start.scss
```diff
@@ -677,6 +677,55 @@ html, body {
     margin-left: 24px;
 }
 
+@media (min-width: 576px) {
+    .modal-dialog {
+        max-width: 575px;
+    }  
+}
```
kinueng: A style suggestion is to move this `@media` into the `.modal-content` on line 686 since the `@media` only applies to the modal.  I don't have a strong preference and just wanted to mention the option since we are using Sass rather than vanilla CSS.

src/main/content/start.html
```diff
@@ -90,11 +90,32 @@ <h2 class="section_title">Create a starter application</h2>
...
+      <div class="modal-content">
+        <div class="modal-header">
+          <button type="button" class="close generate-project-modal-close" data-dismiss="modal" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
+        </div>
+        <div class="modal-body">
+            <h3>Your zip file has been beamed to your computer!</h3>
+            <p>Unzip it and follow the README instructions or run this command.</p>
+            <div id="cmd_to_run" class="cmd_to_run_class"></div>
```
kinueng: A style suggestion is to remove `_class` from `cmd_to_run` since it is verbose.  Functionally it is fine.

kinueng (dismissed):
Thank you for making the updates.  I like that you made the build command in the modal easy to copy to clipboard.  Anything that makes the UX easier is a plus.  

Let's get this into `draft` and allow for others to give us feedback on the new modal.  I would be particularly interested in the design team's feedback since what is implemented deviates from their original design.

I left two code style suggestions on the file diffs.

kinueng:
A screenshot of this PR running in my local `jekyll s` envirohment 
image

src/main/content/_assets/css/start.scss
Sreejith-Websphere: done

src/main/content/start.html
Sreejith-Websphere: done

src/main/content/_assets/css/start.scss
Sreejith-Websphere: the .modal-dialog is parent class of .modal-content so here I changed style belongs to modal under one parent id #generate-project-modal

src/main/content/start.html
Sreejith-Websphere: removed _class

src/main/content/_assets/css/start.scss
kinueng: Nice work

kinueng (approved):
Looks good.  Thank you for the additional changes

PR #2529, opened 2022-02-21
Docs redirect 404

Sreejith-Websphere:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


kinueng (commented):
Is there an opportunity to test this new logic with the Java tests under https://github.com/OpenLiberty/openliberty.io/tree/prod/src/test/java/io/openliberty/website?  

The complaints from the 500 errors are coming from real users trying to use the website.  It would be ideal if we can test how we handle 500 return codes in our automated tests instead of the users.

Sreejith-Websphere:
added test case for 500 return

src/test/java/io/openliberty/website/it/OpenLibertyEndpointIT.java
```diff
@@ -81,6 +85,23 @@ public void testNewReleaseCausesUpdate() throws InterruptedException, IOExceptio
...
+        httpURLConnection.getErrorStream()));
+        String inputLine;
+        StringBuilder response = new StringBuilder();
+        while ((inputLine = in.readLine()) != null) {
+            response.append(inputLine);
+        }
+        in.close();
+        String responseBody = response.toString();
+        assertEquals(500,responseCode);
+        assertTrue(responseBody.contains("com.ibm.ws.webcontainer.exception.IncludeFileNotFoundException") && responseBody.contains("SRVE0190E") && responseBody.contains("/docs/22.0.0.15/overview.html.gz"));
```
kinueng, line 102: I think we want to check if the response is 404 rather than 500.

kinueng: This pull request would need to be in production before this testcase can start checking for FileNotFound HTTP 500 maps to 404 page

kinueng (approved):
The product code changes LGTM.  We may need to revisit the test case later.

Sreejith-Websphere:
ok

PR #2545, opened 2022-03-07
test doc latest version segment

Sreejith-Websphere:
- [x] Firefox (Desktop)
- [x] Safari (Desktop)
- [x] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/scripts/build/parse_features_toc.py
+++ b/scripts/build/parse_features_toc.py
@@ -51,15 +51,15 @@ print(versions)
 
 # Only process the features of the highest version for the docs draft site
 if(os.getenv("DRAFT_SITE") or os.getenv("DOCS_DRAFT_SITE")):
-    max = ['0', '0', '0', '0']
-    for version in versions:
-        nums = version.split('.')
-        if(int(nums[0]) > int(max[0])):
-            max = nums
-        elif(int(nums[0]) == int(max[0]) and int(nums[3]) > int(max[3])):
-            max = nums
-    max = '.'.join(max)
-
+    # max = ['0', '0', '0', '0']
+    # for version in versions:
+    #     nums = version.split('.')
+    #     if(int(nums[0]) > int(max[0])):
+    #         max = nums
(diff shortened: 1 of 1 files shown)


aknguyen7 (commented):
reference the comment out codes

scripts/build/parse_features_toc.py
```diff
@@ -51,15 +51,15 @@ def createVersionHref(parent, tocHref, tocString):
...
-    max = ['0', '0', '0', '0']
-    for version in versions:
-        nums = version.split('.')
-        if(int(nums[0]) > int(max[0])):
-            max = nums
-        elif(int(nums[0]) == int(max[0]) and int(nums[3]) > int(max[3])):
-            max = nums
-    max = '.'.join(max)
-
+    # max = ['0', '0', '0', '0']
```
aknguyen7, line 56: We might want to put a comment on why we comment out this code block
The docs draft playbook specifies to only pull in the docs-generated content for draft branch which is currently 22.0.0.3 whereas the feature script is only looking under 22.0.0.2 (with this current code) that does exist but it doesn’t have the features since draft only build `latest`.

aknguyen7 (commented):
the rest lgtm

scripts/build/parse_features_toc.py
Sreejith-Websphere: Here the commented code Only process the features of the highest version for the docs draft site.If we were using older version of antora here the max value will be come as 22.0.0.3 because that is the highest version now. But with antora update when we add this line in docs-playbook.yml
urls:
  latest_version_segment: latest
the highest version will be automatically convert as latest (eg: here 22.0.0.3 will be coming as latest)
 So since this commented code is only to find highest version that now antora is doing for us and will come as latest

Sreejith-Websphere:
added comments for commented out code

Sreejith-Websphere (commented):
done

steven1046 (approved):

PR #2631, opened 2022-05-02
javadoc url with parameters cannot be switched correctly

Sreejith-Websphere:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/java/io/openliberty/website/RedirectFilter.java
+++ b/src/main/java/io/openliberty/website/RedirectFilter.java
@@ -62,6 +62,10 @@ public class RedirectFilter implements Filter {
         // url, it won't be a direct from and to.
         if (startsWithMatch) {
             String uri = ((HttpServletRequest) req).getRequestURI();
+            String queryString = ((HttpServletRequest) req).getQueryString();
+            if (queryString != null) {
+                uri = uri +"?"+ queryString;
+            }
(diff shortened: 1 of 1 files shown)


aknguyen7 (commented):
for my understanding, the problem only seen in 22.0.0.4 and it works in latest and 22.0.0.3, trying to understand what is it in 22.0.0.4 that would make this different?

aknguyen7 (approved):
lgtm

Sreejith-Websphere:
in draft site just check as follows
load this url
[https://draft-openlibertyio.mybluemix.net/docs/22.0.0.4/reference/javadoc/microprofile-5.0-ja[…]ation/package-frame.html&class=overview-summary.html](https://draft-openlibertyio.mybluemix.net/docs/22.0.0.4/reference/javadoc/microprofile-5.0-javadoc.html?package=jakarta/activation/package-frame.html&class=overview-summary.html)
then instead of 22.0.0.4 type as latest or any other versions like 22.0.0.3,22.0.0.2 etc it will not remove query params
but when you type actual latest version 22.0.0.5 it will convert to latest but it removes query params
this is fixed now

PR #2648, opened 2022-05-12
Update production

### LINKED ISSUE #2612: Text under H3 heading is larger than other text
We recently discovered that in the docs pages, any text under an H3 heading is showing up as weirdly large. For example see 

 https://openliberty.io/docs/latest/network-hardening.html#_mutual_authentication

In this case, **Mutual authentication** is an H3 header, and the text looks large, while **TLS connectivity** is H2 and the text looks normal. Also seems that bulleted text, even under H3, stays the same- as seen under **Mutual authentication** . This problem is showing up anywhere we have H3 text in the docs, including any feature pages with examples, which are all under h3 headiings.

Here is the problem captured in a screenshot 
image

kinueng:

Fixes https://github.com/openliberty/openliberty.io/issues/2612

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/antora_ui/src/css/doc.css
+++ b/src/main/content/antora_ui/src/css/doc.css
@@ -591,7 +591,7 @@ kbd,
   height: 100%;
 }
 
-.doc .paragraph p {
+.noversion-paragraph {
   font-size: .94444rem;
 }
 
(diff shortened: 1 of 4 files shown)


Sreejith-Websphere (approved):
lgtm

PR #2658, opened 2022-05-19
Upgrade node

Sreejith-Websphere:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/scripts/build/docs_part_1.sh
+++ b/scripts/build/docs_part_1.sh
@@ -1,3 +1,18 @@
+#comment this below line in local if nvm already installed
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
+export NVM_DIR="$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")"
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" # This loads nvm
+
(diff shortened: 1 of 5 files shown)


scripts/build/docs_part_1.sh
```diff
@@ -1,11 +1,26 @@
...
+
 set -e
 export BUILD_SCRIPTS_DIR=$(dirname $0)
 echo "BUILD_SCRIPTS_DIR: $BUILD_SCRIPTS_DIR"
 
 #Antora Portion of Docs
 echo "Begin building of Antora portion of docs"
 
-$BUILD_SCRIPTS_DIR/node_install.sh
+#$BUILD_SCRIPTS_DIR/node_install.sh
```
kinueng: Are we effectively saying with this PR that we no longer need the `node_install.sh` file in our automated builds?

I do see that we call `node_install.sh` in other bash scripts, but I think those scripts are for local development only.

Sreejith-Websphere: yes you are right for our automated build we not need to call node install file. The file it calls node install file its for local dev

kinueng:
```suggestion
```

I would suggest deleting the line over commenting it out

Sreejith-Websphere: ok

aknguyen7 (approved):

Issue #2595, opened 2022-04-07
Create container images for openliberty.io


kinueng:
Create container images for openliberty.io.  Have multiple images to accommodate the various site configurations documented here https://github.com/OpenLiberty/openliberty.io/wiki/Websites-&-branch-breakdown

yeekangc:
This should be helpful.  If we can make one available for development purposes (for blog authors), that is useful too?

kinueng:
This work is being done on branch `docker` https://github.com/OpenLiberty/openliberty.io/tree/docker

PR #2717, opened 2022-06-09
Dockerfiles

aknguyen7:
docker files for CE


aknguyen7:
relate to https://github.com/OpenLiberty/openliberty.io/issues/2595

docker/Dockerfile.builder
```diff
@@ -0,0 +1,26 @@
+#
+#
+# One major factor for using Ubuntu 18.04 LTS is to stay on an older Python 3.6 until the python
+# scripts are migrated to use a higher version
```
kinueng:
```suggestion
# One major factor for using `pipeline-base-image:2.6` is to stay on an older Python 3.6 until the python
# scripts are migrated to use a higher version
```

docker/Dockerfile.builder
```diff
@@ -0,0 +1,26 @@
...
+FROM icr.io/continuous-delivery/pipeline/pipeline-base-image:2.6 as builder
+SHELL ["/bin/bash", "-c"]
+
+COPY scripts /scripts
+
+RUN apt-get update && \
+    apt-get -y install curl gnupg build-essential ruby-full git python3 python3-bs4 python3-lxml \
+    nodejs libgdbm-dev libncurses5-dev automake libtool bison libffi-dev
```
kinueng: I see `nodejs` is part of this list. What is the difference between doing `apt-get install nodejs` vs `node_install.sh`?

scripts/build/node_install.sh
```diff
@@ -1,14 +1,23 @@
...
+echo "$PATH"
+echo "node --version"
+node --version
+echo "npm --version"
+npm --version
+echo "ls -g --depth=0"
+npm ls -g --depth=0
 
 timer_end=$(date +%s)
 echo "Total execution time for installing Node: '$(date -u --date @$(( $timer_end - $timer_start )) +%H:%M:%S)'"
```
kinueng:
```suggestion
echo "Total execution time for installing Node: '$(date -u --date @$(( $timer_end - $timer_start )) +%H:%M:%S)'"

```

Adding EOL to file

scripts/build/node_install.sh
```diff
@@ -1,14 +1,23 @@
...
 echo "Install Node"
-curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
-export NVM_DIR="$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")"
-[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" # This loads nvm
-nvm install --lts
-
-echo "npm version:"
-echo `npm -v`
+#comment this below line in local if nvm already installed
+export NVM_DIR=/root/.nvm
```
kinueng, line 7: This seems like something specific to only the Docker build environment. Could we move `export NVM_DIR=/root/.nvm` to the Dockerfile?

aknguyen7: we are using `NVM_DIR` in line 9-12

docker/Dockerfile.builder
aknguyen7: i think they are the same - to install node, will remove `nodejs` part of this list and test it out in demo2

kinueng (approved):

Issue #2720, opened 2022-06-16
Open Graph images for blog posts not displaying on Twitter (but are fine on LinkedIn)


lauracowen:
The open graph images (Twitter image picked up from a blog post) doesn’t seem to be working at the moment.

For example, the recent Tweet of this blog post just uses the default image: https://twitter.com/emilyfhjiang/status/1536280727553187840

A different image has been specified in the source: `open-graph-image: https://openliberty.io/img/blog/deepdive.png`

Looking back, another recent post tweeted in early May is also defaulting the image, even though a different image file is, again, specified in the source header: https://twitter.com/emilyfhjiang/status/1521791420201377792
`open-graph-image: https://openliberty.io/img/blog/blog_microstream_starter.png`

On further investigation by @mbroz2 ...

for some reason it's padded with spaces, so that may be why:
![image](https://user-images.githubusercontent.com/3322532/174104159-2521f2c2-1c04-42c4-be97-56b48bf3699f.png)

and linkedin seems to render it:
![image (1)](https://user-images.githubusercontent.com/3322532/174104215-0742a15b-57d0-4277-85cf-b89b91f9ec80.png)

kinueng:
I agree we should remove the starting and trailing whitespaces in the `og:image` property, however I think there is a bigger problem.

Per Twitter's documentation, they look for twitter metadata before falling back to Open Graph metadata
https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started#opengraph

> When the Twitter card processor looks for tags on a page, it first checks for the Twitter-specific property, and if not present, falls back to the supported Open Graph property.

We currently **always** set the Twitter specific image property on all blog posts
`<meta name="twitter:image" content="https://openliberty.io/img/twitter_card.jpg">`

lauracowen:
Ah....so we're effectively, for Twitter, overwriting our own Open Graph property. Could we automatically set the Twitter specific image property to whatever is set for the Open Graph property?

kinueng:
I decided to remove the `twitter:image` property completely. Since the front matter attribute for our blog Asciidoc files is `open-graph-image`, I didn't want to tie the attribute to the Twitter property. I will trust Twitter will fallback (per their documentation) to the `og:image` property. I will test to confirm.

kinueng:
[...]

PR #2722, opened 2022-06-22
Remove start/end whitespace on `og:image`; Remove `twitter:image`

kinueng:

For https://github.com/openliberty/openliberty.io/issues/2720

Per Twitter's [documentation](https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started#opengraph), Twitter will fallback to using `og:image` if we remove the `twitter:image` property. I am in favor of removing `twitter:image` rather than keeping it in sync with `og:image`. The reason is that the blog post Asciidoc front matter clearly says `open-graph-image` and has no mention of Twitter properties.

> Twitter card tags look similar to Open Graph tags, and are based on the same conventions as the Open Graph protocol. When using Open Graph protocol to describe data on a page, it is easy to generate a Twitter card without duplicating tags and data. When the Twitter card processor looks for tags on a page, it first checks for the Twitter-specific property, and if not present, falls back to the supported Open Graph property. This allows for both to be defined on the page independently, and minimizes the amount of duplicate markup required to describe content and experience.

Tested with commit in `draft` https://github.com/OpenLiberty/openliberty.io/commit/6f711b98975bd87a22644027950ae76d622405b1

Tested default image with blog post
https://draft-openlibertyio.mybluemix.net/blog/2021/09/14/open-liberty-stack-gradle.html

Tested Open Graph image with blog post
https://draft-openlibertyio.mybluemix.net/blog/2021/09/14/open-liberty-stack-gradle.html

Problem with testing...
I could not get the `og:image` to show up when creating a tweet. I suspect it is due to CORS issues when the URL is domain `draft-openlibertyio.mybluemix.net` and the `og:image` is domain `openliberty.io`

# Example of trying to preview tweet with draft site
image

# Example of using an openliberty.io URL
image

I think we have no choice but to test this minor change in production to bypass possible CORS issues.

- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools

### DIFF
--- a/src/main/content/_includes/head.html
+++ b/src/main/content/_includes/head.html
@@ -7,6 +7,8 @@
   {% assign description = site.description | strip_html | normalize_whitespace | truncate: 160 | escape %}
 {% endif %}
 
+{% assign default_image = 'https://openliberty.io/img/twitter_card.jpg' %}
+
 <head>
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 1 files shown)


aknguyen7 (approved):

kinueng:
I learned today that the reason why the image was not showing up for our test sites (e.g. non-openliberyt.io domains) is due to the `robots.txt` on those dev sites.

```ERROR: Fetching the page failed because it's denied by robots.txt.```

PR #2741, opened 2022-07-13
translation initial release with english yml and jekyll plugin

Sreejith-Websphere:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


kinueng:
Let's also include pages like the `guides.html` and `blog.html` in this PR

Sreejith-Websphere:
@kinueng added guides and blog html page using jekyll translation plugin

src/main/content/_i18n/en.yml
```diff
@@ -0,0 +1,281 @@
...
+  prometheus: Prometheus
+  maven: Maven
+  kubernetes: Kubernetes
+  kibana: Kibana
+  elastic_search: Elastic Search
+  follow: Follow
+  and: and
+  and_with: And with
+  or: or
+  trans_yes: Yes`
```
kinueng:
```suggestion
  trans_yes: Yes
```

Is this suppose to have the trailing `? It seems like a typo

src/main/content/_includes/overview.html
```diff
@@ -5,10 +5,10 @@
         {% if page.title == "Server commands overview" %}
             <p id="github_link">
                 <img id="docs_github_logo" src="/img/docs_github_logo.svg" alt="GitHub logo">
-                Contribute on GitHub:
-                <a id="open_issue_link" href="" target="_blank" rel="noopener">Open Issue</a>
+                {% t overview.contribute_on_github %}:
```
kinueng, line 8: Do you think we should include the `:` in the message file? My thinking is that we keep all the message in one place and it is likely translation will handle symbols in the messages without issue.

src/main/content/support.html
```diff
@@ -53,8 +53,8 @@ <h2>WebSphere Application Server</h2>
                             <img id="twitter_logo" src="/img/twitter_logo.svg" class="d-block mx-auto" alt="twitter">
                         </figure>
                         <div class="card-body">
-                            <h5 class="card-title">@openlibertyio</h5>
-                            <p class="card-text">News & updates</p>
+                            <h5 class="card-title">@{% t global.openlibertyio %}</h5>
```
kinueng, line 61: I would prefer to move the `@` into the message file so the message is all together.

src/main/content/index.html
```diff
@@ -37,17 +44,17 @@ <h3 id="intro_lighten_up_title">A lightweight open framework for building fast a
...
-                <img id="phone" class="d-none d-md-block" src="img/home_phone.png" alt="Phone">
-                <img id="mouse" class="d-none d-md-block" src="img/home_mouse.png" alt="Mouse">
+                <img id="person_1" src="/img/home_person_1.png" alt="Person 1">
+                <img id="coffee" src="/img/home_coffee.png" alt="Coffee cup">
+                <img id="shoe" src="/img/home_shoe.png" alt="Shoe">
+                <img id="laptop" class="d-none d-md-block" src="/img/home_laptop.png" alt="Laptop">
+                <img id="ball" class="d-none d-md-block" src="/img/home_ball.png" alt="Ball">
+                <img id="dog" class="d-none d-md-block" src="/img/home_dog.png" alt="Dog">
+                <img id="phone" class="d-none d-md-block" src="/img/home_phone.png" alt="Phone">
+                <img id="mouse" class="d-none d-md-block" src="/img/home_mouse.png" alt="Mouse">
```
kinueng, line 54: Are we making these absolute because the `src` values were broken after we added `openliberty.io/de/...`? I can imagine that the non-english urls were doing `openliberty.io/de/img/...`, which causes the images to be 404.

src/main/content/index.html
```diff
@@ -62,36 +69,76 @@ <h3 id="intro_lighten_up_title">A lightweight open framework for building fast a
...
+                    {% else %}
+                    <div class="green_box"><a href="/{{site.lang}}/guides/microprofile-rest-client.html">{% t global.restclient %}</a></div>
+                    <div class="green_box"><a href="/{{site.lang}}/guides/rest-intro.html">{% t global.jaxrs %}</a></div>
+                    <div class="green_box"><a href="/{{site.lang}}/guides/rest-client-java.html">{% t global.jsonp %}</a></div>
+                    <div class="green_box"><a href="/{{site.lang}}/guides/rest-client-java.html">{% t global.jsonb %}</a></div>
+                    <div class="green_box"><a href="/{{site.lang}}/guides/cdi-intro.html">{% t global.cdi %}</a></div>
+                    {% endif %}
                 </div>
-                <h2 id="table_label" class="d-none d-md-block">MicroProfile</h2>
+                <h2 id="table_label" class="d-none d-md-block">{% t global.microprofile %}</h2>
```
kinueng: My gut feeling is that these strings will not be translated because they are words for a code specification and therefore should not be translated.

src/main/content/index.html
```diff
@@ -152,62 +217,40 @@ <h2 class="develop_title">Monitor microservices in the cloud</h2>
...
-                        <p>Kibana</p>
-                    </a>
-                    <a href="https://www.elastic.co/elasticsearch/">
-                        <p>Elastic Search</p>
-                    </a>
+                    <p>{% t global.prometheus %}</p>
+                    <p>{% t global.maven %}</p>
+                    <p>{% t global.kubernetes %}</p>
+                    <p>{% t global.kibana %}</p>
+                    <p>{% t global.elastic_search %}</p>
```
kinueng: I also think since these are company names, they also may not need to be translated. We might have to keep the hardcoded English string in the HTML to avoid any accidental translation of the brand names.

src/main/content/start.html
```diff
@@ -339,10 +388,11 @@ <h3 class="eclipse_development_tools_title pull-left">Development builds</h3>
...
     </p>
     <p id="more_downloads_text">
-        More downloads are on the way.<br>
-        Check back soon!
+        {% t start.download_package_section.more_downloads_on_the_way %}
+        <br>
+        {% t start.download_package_section.check_back_soon %}
     </p>
-</div>
+</div>
```
kinueng:
```suggestion
</div>

```

Add EOL to this file

kinueng (changes requested):
A general comment about the various places where we are using Liquid `if-else` like below

```
  {% if site.lang == 'en' %}
    <a ...></a>
  {% else %}
    <a ...></a>
  {% endif %}
```

I would prefer we consolidate all of the `if-else` to being a single `if-else` block at the top of the file. This top block will determine what the base URL is for all of the `href`. This will clean up the code from having duplicate HTML elements that only differ by their `href` attribute.

Here is what I got working in my test environment with very light testing in file `index.html`

```
{% assign baseURL = '' %}
{% if site.lang != 'en' %}
    {% assign baseURL = '/' | append: site.lang %}
{% endif %}

...
...
...
<a id="download_link" href="{{baseURL}}/start">{% t home.get_openliberty %}</a>
```

Sreejith-Websphere:
ok thanks @kinueng for the review I will look after these changes

src/main/content/index.html
Sreejith-Websphere: yes

src/main/content/_i18n/en.yml
Sreejith-Websphere: its not typo it should be within single quotes like 'Yes' because I saw when we just add as Yes yaml doesn't interpret, it throws error
when we add within single quotes output looks like below
Screenshot 2022-07-20 at 4 55 23 PM

src/main/content/_includes/overview.html
Sreejith-Websphere: yes but when we use colon : it shows error because yaml interprets key:value pair

src/main/content/support.html
Sreejith-Websphere: when we add to message file it shows below error
atopenlibertyio: @openliberty.io
found character that cannot start any token while scanning for the next token

kinueng (approved):
LGTM

Issue #2484, opened 2022-02-07
Update javadoc canonicals to include query params


mbroz2:
Currently, the canonical points to just the .html file without any query params.  For example, for
[https://openliberty.io/docs/22.0.0.1/reference/javadoc/microprofile-5.0-javadoc.html[…]activation/package-summary.html&package=allclasses-frame.html](https://openliberty.io/docs/22.0.0.1/reference/javadoc/microprofile-5.0-javadoc.html?class=jakarta/activation/package-summary.html&package=allclasses-frame.html)
the canonical is just
https://openliberty.io/docs/22.0.0.1/reference/javadoc/microprofile-5.0-javadoc.html

This means that as far as crawlers are concerned, there is only page per javadoc that should get indexed, instead of all of them.

kinueng:
Google Search Console was scheduled to crawl the Javadoc pages and this issue was discovered.

kinueng:
Please see if Google Search Console is honoring the new query parameters in the canonicals field.  The changes are in production.

mbroz2:
No it doesn't see like doing the canonical update on the client side is affecting results in Search Console (likely they're either grabbing canonical prior to js running, or directly out of the html).  Seems like this will need to be injected at build time or server side at run time.

Also, there seems to be URLs at https://openliberty.io/docs/modules/reference/liberty-jakartaee9.1-javadoc which are not being redirected. For example:
https://openliberty.io/docs/modules/reference/liberty-jakartaee9.1-javadoc/index.html?jakarta/batch/runtime/context/package-summary.html
https://openliberty.io/docs/modules/reference/liberty-jakartaee9.1-javadoc/jakarta/faces/component/NamingContainer.html not getting redirected

steven1046:
I think we need to add Jakarta logic here: https://github.com/OpenLiberty/openliberty.io/blob/prod/src/main/content/_assets/js/javadoc-redirect.js

Sreejith-Websphere:
i created liberty-jakartaee not redirecting as a seperate issue 
 #2574

kinueng:
We will explore creating a new Servlet filter to add the correct canonicals to the HTML content served by a redirect request.

Edit June 14: When trying to add a new Servlet filter, there was a roadblock on getting the new filter to run (e.g. ordering of the filters) correctly. We want the filter to add the query parameter to the HTML that is returned as a response to the redirect HTTP request.

kinueng:
[...]

PR #2764, opened 2022-08-10
setting javadoc canonical url query param

Sreejith-Websphere:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)
- [ ] iOS (Mobile)
- [ ] Android (Mobile)
#### Running validation tools


pom.xml
```diff
@@ -79,6 +79,13 @@
 			<version>5.5.4</version>
 			<scope>test</scope>
 		</dependency>
+
+        <!-- jsoup HTML parser library @ https://jsoup.org/ -->
+		<dependency>
+			<groupId>org.jsoup</groupId>
+			<artifactId>jsoup</artifactId>
+			<version>1.15.2</version>
+		</dependency>
```
kinueng, line 88: I double-checked the code license for this library and saw it was `MIT` at this link https://mvnrepository.com/artifact/org.jsoup/jsoup. Glad to see this is an actively supported dependency too.

src/main/java/io/openliberty/website/TLSFilter.java
```diff
@@ -125,12 +136,30 @@ public void doFilter(ServletRequest req, ServletResponse resp, FilterChain chain
                 }
                 if (doGzip) {
                     response.setHeader("Content-Type", "text/html");
-                    response.setHeader("Content-Encoding", "gzip");
                     doFilter = false;
                     try {
-                        req.getRequestDispatcher(uri.concat(".gz")).include(req, response);
+                        if (queryString != null) {
+                            uriQueryString = uri +"?"+ queryString;
+                            docsPage = Jsoup.connect("https://openliberty.io"+uriQueryString).get();
```
kinueng: We should avoid hardcoding the domain name so we can accommodate our test sites. There should be some methods to access different parts other URL from the request. Maybe `HttpServletRequest.getServerName()`?

src/main/java/io/openliberty/website/TLSFilter.java
```diff
@@ -125,12 +136,30 @@ public void doFilter(ServletRequest req, ServletResponse resp, FilterChain chain
...
+                            String ifError = docsPage.select("head > title").first().text();
+                            if(ifError.contains("404")){
+                                response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+                                response.sendRedirect("/404.html");
+                            }
+                            else {
+                                String updatedCanonicalUrl = uriQueryString.replace(uriQueryString.split("\\/")[2], "latest");
+                                Element result = docsPage.select("link[rel=\"canonical\"]").first();
+                                result.attr("href", "https://openliberty.io"+updatedCanonicalUrl);
```
kinueng: Same comment as before about avoiding hardcoding domain name `openliberty.io`

src/main/java/io/openliberty/website/TLSFilter.java
```diff
@@ -125,12 +136,30 @@ public void doFilter(ServletRequest req, ServletResponse resp, FilterChain chain
...
+                                htmlResponse = response.getWriter();
+                                htmlResponse.println(docsPage.html());
```
kinueng: Do we need to close or flush any kind of stream when using `PrintWriter`? We want to confirm that we do not need to call anything to "close" the writer so we do not have a memory leak that can crash the server.

src/main/java/io/openliberty/website/TLSFilter.java
```diff
@@ -125,12 +136,30 @@ public void doFilter(ServletRequest req, ServletResponse resp, FilterChain chain
...
+                            }
+                        }
+                        else {
+                            response.setHeader("Content-Encoding", "gzip");
+                            req.getRequestDispatcher(uri.concat(".gz")).include(req, response);
+                        }
                     }
-                    catch(FileNotFoundException e) {
+                    catch(FileNotFoundException | NullPointerException e) {
```
kinueng: Can you describe which method calls could throw a NullPointerException (NPE)? I would prefer that we check for nulls in our code and only use the catch for unexpected situations.

Sreejith-Websphere: If we test with any bad format version like 22..0.8 some times it throws NullPointerException with query params

src/main/java/io/openliberty/website/TLSFilter.java
Sreejith-Websphere: @kinueng saw that every where we are setting the canonical url to production url eg: for draft, staging site it points to like eg: "link rel="canonical" href="https://openliberty.io/start/" " that is why I hardcoded above code. Need suggestion on this

Please refer below screenshot where we setting canonical url in antora ui

Screenshot 2022-08-25 at 3 03 29 PM

pom.xml
Sreejith-Websphere: yes

src/main/java/io/openliberty/website/TLSFilter.java
Sreejith-Websphere: added flush and close

src/main/java/io/openliberty/website/TLSFilter.java
kinueng: @Sreejith-Websphere From our last discussion, we would like to remove the hardcoded domain. The test sites are configured with a whole site no index in their `robots.txt` so having `link rel="canonical" href="https://openliberty.io/start/` on a non-openliberty.io domain (i.e. test sites) should not have any adverse effect on how openliberty.io pages are indexed. However, we should replace the hardcoded domain.

We are considered that there will be issues with local development where a the domain includes a port like `https://localhost:9443`. Please see if `HttpServletRequest` or its parent classes have a method that can return the domain regardless of if it has a port or not. For example, the method would return `openliberty.io` or `localhost:4000` with only one method call.

src/main/java/io/openliberty/website/TLSFilter.java
kinueng: I would like to avoid using Java exceptions as a way to handle scenarios that we know can happen. Can you help me find the line that processes the version string? Perhaps we can use a regular expression to validate that there are four integers delimitated by periods and add error handling when the criteria is not met.

For example, in this PR you knew `docsPage.select("head > title").first().text();` can result in a "404".
```
if(ifError.contains("404")){
  response.setStatus(HttpServletResponse.SC_NOT_FOUND);
  response.sendRedirect("/404.html");
}
```

kinueng:
My concern with merging this PR is that we have not diagnosed the root cause of the HTTP 500 that was seen on demo1 in the past. Until we have dedicated focus on really figuring out what caused the 500 error, I am hesitate to move this PR forward.

kinueng:
This may be related to https://github.com/OpenLiberty/openliberty.io/issues/2484

natalie-bernhard:
This work has been completed via Issue #2484 .  Closing this PR.

Issue #2854, opened 2022-09-27
`22.0.0.10` release build not sorted to the top of table


kinueng:
![image](https://user-images.githubusercontent.com/31117513/192594625-e6ada273-e3f6-48a4-9038-f68164ccb85c.png)

NottyCode:
This is caused because the info.json for the releases has the version set in date_time since 22.0.0.2 rather than the release date. The date_time in the json needs to be the date.

kinueng:
Here is the `date_time` in question from `GET` https://openliberty.io/api/builds/data
![image](https://user-images.githubusercontent.com/31117513/192600470-b64cb347-5133-49f9-815e-b17f09625fab.png)

PR #2855, opened 2022-09-27
Update to sort based on version.

NottyCode:
In 22.0.0.2 the uploads were changed from being in a date based directory to a version based one. This messed up the sort logic which assumed that the directory was the date of the release and it means that 22.0.0.10 no longer sorted as the most recent release.

Instead updating so if the version field is set we will sort based on version. The ideal is we will covert each version section to be a three digit version with leading zeros. That means that as long as we don't have 1000 numbers in any component then the sort will be fine. Based on current versioning this will last until the year 3000 or until we do 3 releases a day every day for a year, either seem to be unlikely.

## What was changed and why?
Sorting was shifted from date time to a sort field.

## Link GitHub issue
Issue #2854 

## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)

### DIFF
--- a/src/main/java/io/openliberty/website/data/BuildData.java
+++ b/src/main/java/io/openliberty/website/data/BuildData.java
@@ -68,7 +68,7 @@ public class BuildData {
 
                 @Override
                 public int compare(BuildInfo o1, BuildInfo o2) {
-                    return o2.dateTime.compareTo(o1.dateTime);
+                    return o2.sortField.compareTo(o1.sortField);
                 }
             });
             this.builds.put(type, storedBuilds);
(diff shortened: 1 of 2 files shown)


kinueng:
Testing changes in `draft` branch and corresponding test site through PR https://github.com/OpenLiberty/openliberty.io/pull/2856

kinueng (approved):
Looks good on https://ui-draft-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/start/

> Note: this particular test site does not have blog posts so the missing `Release blog` links are suppose to be missing

image

Issue #2645, opened 2022-05-10
Use minified JavaScript files found on docs page


kinueng:
While running Chrome Lighthouse on this page ([link](https://openliberty.io/docs/latest/reference/javadoc/liberty-jakartaee8-javadoc.html?class=javax/batch/api/chunk/AbstractItemReader.html#close--&package=allclasses-frame.html)), the three JS files below were flagged as potential performance savings if they were minified.

- https://openliberty.io/docs/modules/reference/liberty-javaee8-javadoc/jquery/jszip/dist/jszip.js
~- https://code.jquery.com/jquery-3.3.1.js~
- https://openliberty.io/docs/modules/reference/liberty-javaee8-javadoc/jquery/jquery-1.10.2.js
- https://openliberty.io/docs/modules/reference/liberty-javaee8-javadoc/jquery/jquery-ui.js

![image](https://user-images.githubusercontent.com/31117513/167648849-5964ff36-ccef-445e-bb28-cb277271400c.png)

PR #2868, opened 2022-10-05
Use minified jQuery, use new CDN, use integrity

kinueng:
## What was changed and why?

- Switch to using minified jQuery files for that extra savings on network traffic payload size.
- Switch to a CDN that provides the `integrity` attribute support.

#2869 used to test on draft site https://draft-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/

## Link GitHub issue
Issue #2645

## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)


kinueng:
General take is that there is a performance cost with the new CDN to use the `integrity` attribute. I think its worth the hit for the added check.

Comparing the two CDNs, I have noticed:
- `cdnjs.cloudflare.com` with the additional `integrity` attribute has a slower network end to end time than `code.jquery.com`
- The file size has been reduced from `80.5 kB` to `28.1kB`
- The file compression went from `gzip` to `br` with the new CDN
image

natalie-bernhard (approved):

kinueng:
According to https://tools.pingdom.com/ 
- The homepage - JavaScript went from `844.8 KB` to `771.8 KB`

PR #2870, opened 2022-10-05
Update production

### LINKED ISSUE #2126: Language selector for blog posts
Language selector (we can also set default to what the browser returns, but that's 'free')
(considered a hard prereq as without UX would degrade significantly)
**Size: 1.5pw**

### LINKED ISSUE #2166: Blogs: text in bulleted lists is smaller than body text
I only see this problem in the blogs. Text that is part of an unordered list (bullets) is several pts smaller than the rest of the body text on the page. This has the effect of making bulleted content less prominent and harder to scan. This does not happen with numbered lists, just bullets. Example from https://www.openliberty.io/blog/2021/06/04/database-ssl-primer.html

Screen Shot 2021-06-16 at 10 14 32 AM

### LINKED ISSUE #2830: Two scroll bars initially appear when loading the landing page
When initially loading the home page, two scroll bars appear on the right rather than one, then the second one disappears (occurs on multiple browsers)

kinueng:
## Link GitHub issue
Fixes #2166 
Fixes https://github.com/OpenLiberty/openliberty.io/issues/2830
Fixes https://github.com/OpenLiberty/openliberty.io/issues/2126

## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)

### DIFF
--- a/scripts/build/clone_blogs.sh
+++ b/scripts/build/clone_blogs.sh
@@ -24,7 +24,7 @@ fi
 
 echo "Cloning the $BRANCH_NAME branch of blogs repository..."
 
-git clone https://github.com/OpenLiberty/blogs.git --branch "demo1" blogs_temp
+git clone https://github.com/OpenLiberty/blogs.git --branch $BRANCH_NAME blogs_temp
 
 # This section is moving the blog posts around for the Jekyll build process
 mv blogs_temp/posts/ .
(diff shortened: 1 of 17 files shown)


natalie-bernhard:
Looks good to me

natalie-bernhard (approved):
Looks good to me

Issue #2285, opened 2021-09-28
Archive old releases


mbroz2:
Currently, the /start/ page shows all releases of OL.  We'll want to only show the most recent X releases, same as docs. (X needing definition, likely either all releases in the last 2 years, or a subset where we keep all releases in the last year, plus those from the previous year where X mod 3 = 0.  Another option is to align with Liberty's sec vul iFix statement).

We'd then have some sort of button/toggle to view 'archived' releases.

One implementation would be to show the X most recent ones in the table, and then have a button "Show archived releases" which expands the table to show the older releases.  This is likely the cheapest implementation wise, and does a good job 'hiding' the archived releases.
Another implementation would be the same as above, except the archived releases would be in its own table below the table of regular releases.
Another implementation would be a separate page for the archived releases.

PR #2871, opened 2022-10-05
2285 archive old releases

natalie-bernhard:
## What was changed and why?
Releases older than two years old are now removed from the runtime release table.  A link was added at the bottom of the table that opens a tab to the public DHE where archived releases can be downloaded from.

## Link GitHub issue
Issue #2285 

## Tested using browser:
- [X] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)

### DIFF
--- a/src/main/content/_assets/css/start.scss
+++ b/src/main/content/_assets/css/start.scss
@@ -507,6 +507,10 @@ html, body {
     margin-bottom: 41px;
 }
 
+.archive_sentence {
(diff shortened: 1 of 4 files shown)


natalie-bernhard:
This is ready for review.  Changes to builds.js on lines 600 and 610 makes it so the code follows ES5 rather than ES6.

kinueng (approved):

PR #2872, opened 2022-10-10
2166 blog bulletlist font

natalie-bernhard:
## What was changed and why?
Made the font size of unordered lists on blog posts the same as the font size on the rest of the post.

## Link GitHub issue
Issue #2166 

## Tested using browser:
- [X] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)

### DIFF
--- a/src/main/content/_assets/css/post.scss
+++ b/src/main/content/_assets/css/post.scss
@@ -93,6 +93,7 @@ body {
 #article_body li > p {
     margin-top: 13px;
     margin-bottom: 13px;
+    font-size: 16px;
 }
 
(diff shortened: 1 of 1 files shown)


kinueng (approved):
LGTM.

It is good that we put this change in `post.scss` so the blast radius of the change is only to viewing a single blog post.

I went down a rabbit hole and looked for the original change that cause the issue we are addressing here. https://github.com/OpenLiberty/openliberty.io/pull/2082. The PR doesn't have an issue linked so I do not know what the motivation was for loosing the CSS selector.

PR #2876, opened 2022-10-11
2830 double scrollbar bug

natalie-bernhard:
## What was changed and why?
Overflow CSS for the home page was modified so that one scrollbar renders from the beginning instead of having two scrollbars render initially and then consolidate to one.

## Link GitHub issue
Issue #2830 

## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)

### DIFF
--- a/src/main/content/_assets/css/home.scss
+++ b/src/main/content/_assets/css/home.scss
@@ -15,7 +15,7 @@ html {
 
 body {
   position: relative;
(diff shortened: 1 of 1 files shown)


natalie-bernhard:
Now only one scroll bar renders when loading the draft site

kinueng (approved):

PR #2933, opened 2022-11-30
2909 more downloads logic

### LINKED ISSUE #2909: "More downloads are on the way" at bottom of /start should only appear for an empty or partially rendered table
Currently, "More downloads are on the way" at bottom of /start seems to always appear.  The image should only appear if the table is empty or partially rendered.
image

natalie-bernhard:
## What was changed and why?
Logic was added so that the "More downloads on the way" animation only appears if the request to the DHE fails or if there are less than 24 releases in the downloads table.

## Link GitHub issue
Issue #2909 

## Tested using browser:
- [X] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)

### DIFF
--- a/src/main/content/_assets/js/builds.js
+++ b/src/main/content/_assets/js/builds.js
@@ -1462,10 +1462,13 @@ $(document).ready(function () {
             startAnimation();
         }
 
-        if(!failed_builds_request){
-            $('#bottom_images_container').hide();
-        } else {
+        // if builds only partially render or don't render at all, show the animation
+        var rendered_builds = $("#runtime_releases_table > tbody > tr").length;
+
+        if(rendered_builds < 24 || failed_builds_request){
             $('#bottom_images_container').show();
+        } else{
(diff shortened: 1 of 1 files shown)


natalie-bernhard:
I have attempted to rebase, cherry-pick, and other methods of fixing the commits multiple times but cannot seem to remove these additional earlier commits from the request (the changes these commits make are already present in draft and the feature branch); I think the issue is that these specific commits were merged into prod but not into draft.  I believe a PR into draft from prod (or merging this PR as is) would fix this issue, but I am not sure if this would cause any other issues.

natalie-bernhard:
The "two dot comparison" [here](https://github.com/OpenLiberty/openliberty.io/compare/draft..2909-more-downloads-logic) shows that the only actual differences are the changes made for the feature/bug.

steven1046 (approved):

Issue #2934, opened 2022-12-02
Blog posts aren't reliably syndicating to the Jakarta EE blog


lauracowen:
I count 15 blog posts under the `Jakarta EE` tag for 2022 at: https://openliberty.io/blog/?search=Jakarta_EE&key=tag

However, only two of those posts are syndicated at Jakarta EE blog: https://jakarta.ee/blogs/
https://openliberty.io/blog/2022/11/04/liberty-tools-eclipse-jakarta-ls.html
https://openliberty.io/blog/2022/10/27/22.0.0.12-beta.html

![image](https://user-images.githubusercontent.com/3322532/205330109-6fbe5fc5-422d-4612-b841-232cd46fd3f5.png)

kinueng:
Here is openliberty.io's RSS feed for Jakarta EE blog posts
https://openliberty.io/jakartaee.xml

Oddly, there are more than 2 blog posts in the RSS feed.

kinueng:
I suspect the XML for the Jakarta EE RSS XML contains some syntax issues
https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fopenliberty.io%2Fjakartaee.xml

Our default blog feed is fine
https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fopenliberty.io%2Ffeed.xml

kinueng:
Thought: Is there something we can do to automatically verify our feed.xml files as this is not the first time we had a bug in this area

natalie-bernhard:
Update 12/14/22: Removing GUID/making the GUID unique makes the feed valid again. GUIDs were not guaranteed to be unique due to redirected blogs having the same site.url.  Next step is to figure out if GUID is needed, and if it is, what do we want to use for the GUID?

lauracowen:
What's the GUID for? Where's it used?

natalie-bernhard:
It's used to give each item in the feed a unique identifier/ID. I'm not sure if the Jakarta EE blog uses this parameter/attribute when rendering the posts from the feed, hence why it can't just be removed.  

Currently our feed still produces all of the Jakarta EE posts, so I'm wondering whether the Jakarta EE website is having it's own issues or choosing to limit how many blog posts to render from the feed.

lauracowen:
cc @NottyCode In case you have any insights into this, as you set up the original aggregation?

kinueng:
[...]

PR #2971, opened 2023-01-10
2934 jakarta blog feed

natalie-bernhard:
## What was changed and why?
GUID value is now unique for each post on the feed, so the feed is now valid

## Link GitHub issue
Issue #2934 

## Tested using browser:
- [X] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


kinueng (approved):
This fix seems correct based on my doing some self learning on `guid` in RSS feeds in the links below
- https://validator.w3.org/feed/docs/warning/MissingGuid.html
- https://www.w3schools.com/xml/rss_tag_guid.asp

It looks that this fix is not guaranteed to fix the syndication issue, but we should at least have the jakartaee.xml be a valid format.

PR #3034, opened 2023-02-13
2885 accessible code block

### LINKED ISSUE #2885: copy code block buttons are not accessible to screen readers
While doing JAWS testing on the Open Liberty docs, I found that the "copy code block" buttons are not discoverable by the screen reader. 

Using the "B" key in the JAWS program takes you through all the buttons on the current screen. On multiple pages, I found that this command never finds the code copy buttons. Also, If you hover the over the buttons, the screen reader reads the source file designation for the code example (eg "XML") rather than the alt text for the copy code button. 

It appears there is no way that someone using  a screen reader would know the copy code buttons exist.

tested multiple pages/ code blocks on chrome on a mac

natalie-bernhard:
## What was changed and why?
`Copy code block` button programming was changed to make sure the button was accessible by a screenreader.  Macs have a built-in screenreader known as [VoiceOver](https://support.apple.com/guide/voiceover/welcome/mac) that can be used for testing (Shortcut: Cmd + F5).  Used the screenreader on my local instance and was able to navigate through without missing the `Copy code block` button.

## Link GitHub issue
Issue #2885 

## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


src/main/content/_assets/js/guide-common.js
```diff
@@ -453,6 +453,15 @@ $(document).ready(function () {
         $(this).css("opacity", "1");
     });
 
+    // set up html for copy code block accessibility
+    var static_codeblock_selector = "#guide_content pre:not(.no_copy pre):not(.code_command pre):not(.hotspot pre):not(.code_column pre), #guide_content .code_command .content"
+    $(static_codeblock_selector).each(function (){
+        $(this).wrap('<div class="code_block_wrapper" title="Code block"></div>');  
+    })
+    $('.code_block_wrapper').each(function (){
+        $(this).prepend('<div id="copied_confirmation">Copied to clipboard</div><input type="image" id="copy_to_clipboard" src="/img/guides_copy_button.svg" alt="Copy code block" title="Copy code block"/>');
```
steven1046: Can we change the id of copy_to_clipboard to a class, since there could be many instances of the same id, especially on the start page?

steven1046 (changes requested):

steven1046 (approved):

PR #3164, opened 2023-05-15
3111 Fixes for Cypress Tests to upgrade cypress and fix tests

### LINKED ISSUE #3111: Automate Cypress tests for openliberty.io
As we learn from https://github.com/OpenLiberty/start.openliberty.io/issues/146, we should also automate the existing Cypress tests for openliberty.io found here https://github.com/OpenLiberty/openliberty.io/tree/prod/src/test/cypressjs/cypress/integration

amyreit:
## What was changed and why?
Upgrading cypress level and updating tests to run successfully manually
## Tested using browser:
- [x] Chrome (Desktop)


natalie-bernhard (dismissed):

src/test/cypressjs/README.md
```diff
@@ -4,13 +4,18 @@
...
+- `npx cypress open` to run in the ui
+- `npx cypress run --spec "cypress/e2e/???.cy.js"` to run headless
+
+NOTE - the testOLStarterAllGradle.cy.js and cypress/e2e/testOLMavenAllGradle.cy.js tests use a 
+lot of system memory and run more reliably headless
 
-## Step 3:  Launch the Cypress test runner
-- **Option 1 from step 2** using `cypress.env.json`
-  - `npx cypress open`
-- **Option 2 from step 2** using CLI variable instead of `cypress.env.json`
```
aknguyen7, line 15: is the Option 2 does not work anymore that we're removing this?

amyreit: the cypress.env.json has been removed with new cypress. We have the same strings in cypress.config.js and we could still allow command line but it really doesn't hurt anything if someone accidentally checks in the config file with their values since everyone will have to modify it locally to run anyway - it's not a file that is used by the build process and it just simplified things to have just one way to run it

aknguyen7 (approved):

PR #3277, opened 2023-07-10
Dependabot/npm and yarn/src/main/content/antora UI/tough cookie and node sass  removed

natalie-bernhard:
## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


amyreit (commented):
I thought about this a bit - I am on the fence about where we want the action to run. It would be best to allow the user to choose staging or production, but I'd need to add that option. In reality, I'd want to know if the staging site is broken before it gets pushed but also would be valuable to be able to run it on production just as a sanity check. Not sure if where we want to leave it until I can get around to adding the options. But right now this change is already in prod.

amyreit (commented):
Nevermind

amyreit (approved):

Issue #1550, opened 2020-02-11
Add Open Liberty API & SPI Javadocs


mbroz2:
Currently we only publish Javadocs for MP & JEE; also publish Open Liberty API & SPI Javadocs.

dmuelle:
@hlhoots has been working on this and provided this OL javadoc jar, based on the WebSphere Liberty javadoc. Can we look into adding this to draft to see how it works?

[com.ibm.openliberty.javadoc.zip](https://github.com/OpenLiberty/openliberty.io/files/9473435/com.ibm.openliberty.javadoc.zip)

kinueng:
The documentation in the jar in https://github.com/OpenLiberty/openliberty.io/issues/1550#issuecomment-1234785822 is not formatted in a way where it can be dropped into openliberty.io. There will need to be work in the generation code.

From my recollection, the jar is formatted like these docs here https://www.ibm.com/docs/en/was-liberty/nd?topic=SSAW57_liberty/com.ibm.websphere.javadoc.liberty.doc/api_ndz.htm

hlhoots:
@kinueng The link you pasted on the comment above - where does that content come from?

dmuelle:
Hi @hlhoots those HTML files are generated with the Liberty GM build for each release. I get them from [a link that @samwatibm posts to #was-dev-announcements monthly ](https://ibm-cloud.slack.com/archives/C03HP6CHPPD/p1670344740902969). Here's where I got the files for 22.0.0.13:
https://rtpgsa.ibm.com/gsa/rtpgsa/projects/l/liberty.shipped/liberty/WASX.LIBERTY/shipped/cl221320221205-1900.59.linux/linux/zipper/externals/installables/

On that page, the link for [com.ibm.websphere.javadoc.liberty.doc.jar](https://rtpgsa.ibm.com/gsa/rtpgsa/projects/l/liberty.shipped/liberty/WASX.LIBERTY/shipped/cl221320221205-1900.59.linux/linux/zipper/externals/installables/com.ibm.websphere.javadoc.liberty.doc.jar) downloads the JAR of html files for the javadoc, which I then manually upload to IBM Docs cloud object storage. 

[...]

PR #3278, opened 2023-07-11
Update Dockerfile.staging - Test skipping war bundling step

kinueng:
The container images already use the exploded war. Skip over the war bundling step.

## What was changed and why?
For #1550 

Tested using `ui-only` container image on local machine. Website renders fine.

## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)


aknguyen7 (approved):
LGTM

PR #3331, opened 2023-08-23
Add git to docker images

steven1046:
## What was changed and why?
Sometimes the build fails due to git not installed in the later image when trying to run `RUN configure.sh`. This should fix that. `apt-get clean` cleans up temporary files used to install packages.

## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


natalie-bernhard (approved):

steven1046:
Closing since there is still some more debugging to do, some of the images don't have apt-get.

PR #3345, opened 2023-09-05
Lowercase 'Eclipse developer tools'

scottkurz:
## What was changed and why?

This doesn't seem like it should be capitalized.  We had names like: 
Open Liberty Tools, WebSphere Developer Tools, etc.
but we never had an "Eclipse Developer Tools" right?

@yeekangc would you agree?

I didn't do any testing by the way, just find/replace.

### DIFF
--- a/src/main/content/_i18n/en.yml
+++ b/src/main/content/_i18n/en.yml
@@ -114,7 +114,7 @@ start:
     eclipse_link: Eclipse IDE
     intellij_link: IntelliJ IDEA
     vscode_link: Visual Studio Code
-    section_text3: The classic Eclipse Developer Tools can be found in
+    section_text3: The classic Eclipse developer tools can be found in
     public_dhe_link: 'this repository, '
     section_text4: and you can find
     github_link: instructions on how to install the classic tools here.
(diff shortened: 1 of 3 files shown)


steven1046:
I see on https://marketplace.eclipse.org/content/eclipse-web-developer-tools that they are capitalized. @cherylking what do you think?

cherylking:
@steven1046 I agree with Scott.

steven1046 (dismissed):

steven1046 (approved):

steven1046:
We'll need to merge into draft, then the same branch into staging, and then the website team will deliver staging->prod.

steven1046:
Scott could you please open a pull request from scottkurz:lowercase-eclipse-devtools to staging?

scottkurz:
> I see on https://marketplace.eclipse.org/content/eclipse-web-developer-tools that they are capitalized. 

True, the "Eclipse Web Developer Tools" are the dependencies or base of: Open Liberty Tools + WebSphere Developer Tools, but I think in this context we're describing the IBM/Liberty add-ons, not the base, so lowercasing is better.

(I opened the staging PR too BTW)

PR #3353, opened 2023-09-07
remove workaround from prod

natalie-bernhard:
## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


steven1046 (approved):

steven1046:
Approved with the understanding that we'll merge it into staging as well, but we tested this on draft

PR #3423, opened 2023-10-13
Update prod

steven1046:
## What was changed and why?
* Add accessibility check on PRs
* Add anchors to /start
* Update the starter to choose Java 11 when MicroProfile 6 or Jakarta 10 are chosen and the Java SE version was lower than 11 - making it smarter and not always choosing the first option in the Java SE list so when we add future Java SE versions it won't choose the highest.
* Change "Eclipse Developer Tools" to "Eclipse developer tools"

## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [x] Chrome (Desktop)


natalie-bernhard (approved):
Looks good to me

PR #3444, opened 2023-10-17
Added latest blog with release and latest blog without release in home page

SteveSamJacob19:
## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


steven1046:
It might be because the latest blog is just a draft blog, but we need to make sure on staging that the 2 blog cards are the same height:
https://draft-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/
image

steven1046 (approved):

PR #3487, opened 2023-11-14
3427 accessibility /blog

navaneethsnair1:
## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


src/main/content/_assets/js/post.js
```diff
@@ -9,6 +9,11 @@
  *     IBM Corporation - initial API and implementation
  *******************************************************************************/
 
+const preTags = document.querySelectorAll('pre');
+preTags.forEach(preTag => {
```
natalie-bernhard: Since we currently use ES5 (a certain variety of JS), we cannot use the .forEach function.  Can you please modify this to be a standard for loop?

natalie-bernhard (changes requested):

natalie-bernhard (approved):

PR #3494, opened 2023-11-20
Fixed appearance of TOC in interactive guides

### LINKED ISSUE #3463: Interactive Guides table of contents isn't fully appearing in Chrome
https://openliberty.io/guides/microprofile-config-intro.html
As you can see the toc is hidden somewhat on the initial load:
image

SteveSamJacob19:
The interactive guides initially loads up with end of window shown and hence the TOC gets its top value updated to negative number. On setting a delay of 1 sec in triggering the function, this issue was resolved.


natalie-bernhard (changes requested):
Ideally, we'd like to avoid solutions that rely on a timeout, especially since this function applies to all pages on the site.  Can you try to find a different solution that does not rely on a timeout?

PR #3543, opened 2024-01-03
Revert "Revert "Upgrade to 23.0.0.12 Open Liberty driver (draft)""

natalie-bernhard:
Reverts OpenLiberty/openliberty.io#3537


natalie-bernhard:
@kinueng this is the new PR for draft; when I had to do the package updates yesterday, this needed to be reverted as one issue I saw in the logs was that the `configure.sh` script failed, which was remedied once I initially reverted the OL driver upgrade.  Sorry for any confusion.

PR #3596, opened 2024-02-01
Made Blog author name with no Github ID unclickable

### LINKED ISSUE #3565: Make blog authors with no Github ID unclickable on blog feed
In the blog feed (/blog), authors with no Github ID are still able to be clicked, which brings the user back to the top of the feed.  We should make the author's name unclickable if they do not have a Github ID, as is already the case when you navigate to a specific post.

SteveSamJacob19:
## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)

### DIFF
--- a/src/main/content/_layouts/post.html
+++ b/src/main/content/_layouts/post.html
@@ -47,7 +47,11 @@ js: post
                             {% endif %}
                         {% endfor %}
                     </div>
-                    <a class="author_name" href="{{ page.author_github }}" target="_blank" rel="noopener">{{ page.author }}</a>
+                    {% if page.author_github %}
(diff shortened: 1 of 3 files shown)


natalie-bernhard:
https://draft-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/zh-Hans/blog/ this doesn't seem to be working on the draft site.  If you scroll to the last blog post in the feed, the second author is still a clickable link that leads to the current page URL.  See about checking the string length instead of something like `if(exists)` as it seems that on inspection, the href is blank even though the if statement shouldn't allow that.

PR #3601, opened 2024-02-05
Minify html before gzip

natalie-bernhard:
## What was changed and why?
Add minification step to build in order to reduce overall size of the website
Other files shown as modified but they aren't, see the two dot comparison here: https://github.com/OpenLiberty/openliberty.io/compare/staging..html-compression
## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


steven1046 (approved):
LGTM!

PR #3620, opened 2024-02-15
Blog authors without github id made unclickable

### LINKED ISSUE #3565: Make blog authors with no Github ID unclickable on blog feed

SteveSamJacob19:
## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


natalie-bernhard (approved):
Looks good to me

PR #3697, opened 2024-04-03
Fixed cutting of toc content in Docs

### LINKED ISSUE #3519: Docs table of contents cut off for very long titles
For example, https://openliberty.io/docs/latest/reference/diff/mp-60-61-diff.html, the table of contents only says `Differences between MicroProfile 6.1` and is cut off from saying `Differences between MicroProfile 6.1 and 6.0`.

SteveSamJacob19:
## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)


natalie-bernhard (changes requested):
This works for the majority of the TOC, but lines under the API/SPI docs are still being cut off.  Could you add a solution for that as well? I will keep these changes on draft in the meantime since they do not affect functionality.

PR #3737, opened 2024-05-15
automate plugin versions displayed. 

### LINKED ISSUE #3346: Update plugin versions under `Add to an existing application` programatically
Use an API request to get the latest maven/gradle plugin versions for the `Add to an existing application` section on the start page.

navaneethsnair1:
## What was changed and why?
As requested, code from start.openliberty.io is migrated to ol repo.
## Tested using browser:
- [ ] Firefox (Desktop)
- [ ] Safari (Desktop)
- [ ] Chrome (Desktop)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,12 @@
 			<version>5.5.4</version>
 			<scope>test</scope>
 		</dependency>
+
+		<dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+            <version>20240303</version>
+        </dependency>
 	</dependencies>
 
 	<build>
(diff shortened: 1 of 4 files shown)


natalie-bernhard (changes requested):
Code looks overall great! In `start.html`, can you just add in the current latest Maven plugin (3.10.3) and Gradle plugin (3.8.3) versions back as the default text in the code blocks to act as placeholders, just in case the API request fails for any reason.  This will help prevent users from copying an unusable code snippet.  Instead, it may be outdated but should still work even if the version is not completely up to date.

natalie-bernhard (approved):

